PR #16, opened 2023-09-18
Move sync tidy check to the CI only

joaopalet:


vicentepinto98 (commented):
After thinking a bit more about it, I think we should keep the check as part of the `make lint` command and rather only throw error for a diff in the `go.mod` or `go.sum` files. 

The reason is that as a developer I would like to run `make lint` locally and get an error if my dependencies are not synced. And when `make lint` succeeds I'm not expecting it to fail in the CI.

For the diff check, I found this: 
https://stackoverflow.com/questions/8554776/how-to-filter-git-diff-based-on-file-extensions

What do you think?

joaopalet:
> After thinking a bit more about it, I think we should keep the check as part of the `make lint` command and rather only throw error for a diff in the `go.mod` or `go.sum` files.
> 
> The reason is that as a developer I would like to run `make lint` locally and get an error if my dependencies are not synced. And when `make lint` succeeds I'm not expecting it to fail in the CI.
> 
> For the diff check, I found this: https://stackoverflow.com/questions/8554776/how-to-filter-git-diff-based-on-file-extensions
> 
> What do you think?

But the `sync-tidy` script actually syncs and tidies the `go.mod` files by really updating them. So after running it locally the project will always be synced either way.

That's why I think it makes sense to only fail in the CI (as there we don't wan't to commit the changes behind the curtains, and instead fail and warn us that we should run it locally)

.github/workflows/ci.yaml
```diff
@@ -22,6 +22,10 @@ jobs:
         run: make project-tools
       - name: Lint
         if: ${{ matrix.go-version == '1.19' || matrix.go-version == '1.20' }}
-        run: make lint
+        run: |
```
vicentepinto98, line 25: I think it would be better here if we executed `make lint` because other steps might be added later besides the `sync-tidy` and `golangci-lint`. 

But I see the point of failing early after the sync tidy. Maybe moving the `make sync-tidy` step inside the `check-sync-tidy` script and then executing `make lint`, with the drawback of doing the sync twice? :thinking:

vicentepinto98 (changes requested):

vicentepinto98 (approved):

PR #93, opened 2023-09-28
Support key flow authentication

vicentepinto98:


core/clients/key_flow.go
```diff
@@ -0,0 +1,359 @@
...
+const (
+	// Service Account Key Flow
+	// Auth flow env variables
+	ServiceAccountKey     = "STACKIT_SERVICE_ACCOUNT_KEY"
+	PrivateKey            = "STACKIT_PRIVATE_KEY"
+	ServiceAccountKeyPath = "STACKIT_SERVICE_ACCOUNT_KEY_PATH"
+	PrivateKeyPath        = "STACKIT_PRIVATE_KEY_PATH"
+)
+
+var tokenAPI = "https://api.stackit.cloud/service-account/token" //nolint:gosec // linter false positive
```
hcsa73, line 30: What does the linter say?

hcsa73: Also, is it ok, to have the link exposed like this?

core/clients/key_flow.go
```diff
@@ -0,0 +1,359 @@
...
+		tempErr := res.Body.Close()
+		if tempErr != nil {
+			jwks = nil
+			err = fmt.Errorf("closing get jwks response: %w", tempErr)
+		}
+	}()
+	if res.StatusCode == 200 {
+		return io.ReadAll(res.Body)
+	}
+	return nil, fmt.Errorf("error: %s", res.Status)
```
hcsa73, line 358: I'd remove the `error:` part, if you've got nothing to add, just write the status

Also, to be more Go-ish, swap the if condition to be `if res.StatusCode != 200`

core/clients/key_flow.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+// validateToken returns true if token is valid
+func (c *KeyFlow) validateToken(token string) (bool, error) {
+	if token == "" {
+		return false, nil
+	}
+	if _, err := c.parseToken(token); err != nil {
+		if strings.Contains(err.Error(), "401") {
+			c.token = new(TokenResponseBody)
```
hcsa73, line 315: Why is this here?

core/auth/auth.go
```diff
@@ -12,18 +12,27 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/core/config"
 )
 
-const credentialsFilePath = ".stackit/credentials.json" //nolint:gosec // linter false positive
+const (
+	credentialsFilePath             = ".stackit/credentials.json" //nolint:gosec // linter false positive
```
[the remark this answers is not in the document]
vicentepinto98: This is the rule that is triggered: https://securego.io/docs/rules/g101

In this case, this is just a path, so it is a false positive

core/clients/key_flow.go
vicentepinto98: This is the rule that is triggered: https://securego.io/docs/rules/g101

We need to have this url by default for convenience of the user. I think there is no problem in exposing it, but I will double check

core/auth/auth.go
```diff
@@ -12,18 +12,27 @@ import (
...
 type Credentials struct {
 	STACKIT_SERVICE_ACCOUNT_EMAIL string
 	STACKIT_SERVICE_ACCOUNT_TOKEN string
+	STACKIT_SERVICE_ACCOUNT_KEY   string
+	STACKIT_PRIVATE_KEY           string
 }
 
 // SetupAuth sets up authentication based on the configuration. The different options are
 // custom authentication, no authentication, explicit token flow or default authentication
 func SetupAuth(cfg *config.Configuration) (rt http.RoundTripper, err error) {
```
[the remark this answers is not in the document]
vicentepinto98: In my opinion, one of the nicest features of the SDK is that you can just set your environment and then use default authentication for all clients. Having to explicitly configure which authentication you want is less convenient.

Also, the token flow is less secure and will be deprecated at some point, so we should always try to use the key flow as first option. If the user has set the service account key, there should be no reason to not use it. 

However, if for some reason the user really wants to use the token, the option `config.WithToken("TOKEN")` can be used and the token flow will take precedence.

core/clients/key_flow.go
vicentepinto98: `parseToken`  calls `getJwksJSON` which makes a request to get jwks from the service accounts API jwks endpoint using the token. If this call returns 401 unauthorized it means that the token that is being parsed was not issued by the Service Accounts API so the client token field (containing `AccessToken` and `RefreshToken`) should be reset

core/auth/auth.go
```diff
@@ -139,9 +191,34 @@ func readCredentialsFile(path string) (*Credentials, error) {
 	return &credentials, nil
 }
 
+// readCredential reads the specified credentialType from Credentials and returns it as a string
+func readCredential(credentialType string, credentials *Credentials) (string, error) {
+	var credentialValue string
+	switch credentialType {
```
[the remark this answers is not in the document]
vicentepinto98: done

core/auth/auth.go
```diff
@@ -51,36 +66,29 @@ func SetupAuth(cfg *config.Configuration) (rt http.RoundTripper, err error) {
...
-		token, err = readTokenFromCredentialsFile(cfg.CredentialsFilePath)
+	// Key flow
+	rt, err = KeyAuth(cfg)
+	if err != nil {
+		keyFlowErr := err
+		// Token flow
+		rt, err = TokenAuth(cfg)
 		if err != nil {
-			return nil, fmt.Errorf("STACKIT_SERVICE_ACCOUNT_TOKEN not set. Trying to read from credentials file: %w", err)
+			return nil, fmt.Errorf("no valid credentials were found: trying key flow: %s, trying token flow: %w", keyFlowErr.Error(), err)
```
hcsa73, line 89: Can you use "%w" instead of "%s", and have `keyFlowErr` instead of `keyFlowErr.Error()`?

core/auth/auth_test.go
hcsa73, line 1: Shouldn't `TestDefaultAuth` be extended, to check that key flow takes precedence over token flow? I think you can distinguish by checking if the returned `http.RoundTripper` is of type `clients.KeyFlow` or `clients.TokenFlow`

hcsa73 (approved):

core/auth/auth.go
vicentepinto98: No, it throws a compilation error:
`auth/auth.go:89: fmt.Errorf call has more than one error-wrapping directive %w`

hcsa73 (approved):

hcsa73 (approved):

PR #98, opened 2023-10-03
Services - move waiter to package

hcsa73:


CONTRIBUTION.md
```diff
@@ -8,7 +8,7 @@ Your contribution is welcome! Thank you for your interest in contributing to the
 
 ## Developer Guide
 ### Repository structure
-The SDK STACKIT service modules are located under `services`, which are automatically generated from the [REST API specs](https://github.com/stackitcloud/stackit-api-specifications), except for the `wait.go` and `wait_test.go` files. Therefore, changes to these modules besides the `wait.go` and `wait_test.go` files will not be accepted. Instead, consider proposing changes to the generation process in the [Generator repository](https://github.com/stackitcloud/stackit-sdk-generator).
+The SDK STACKIT service modules are located under `services`. The files located in `services/[service]` are are automatically generated from the [REST API specs](https://github.com/stackitcloud/stackit-api-specifications), whereas the ones located in subfolders (like `wait`) are manually maintained. Therefore, changes to files located in `services/[service]` will not be accepted. Instead, consider proposing changes to the generation process in the [Generator repository](https://github.com/stackitcloud/stackit-sdk-generator).
```
vicentepinto98, line 11:
```suggestion
The SDK STACKIT service modules are located under `services`. The files located in `services/[service]` are automatically generated from the [REST API specs](https://github.com/stackitcloud/stackit-api-specifications), whereas the ones located in subfolders (like `wait`) are manually maintained. Therefore, changes to files located in `services/[service]` will not be accepted. Instead, consider proposing changes to the generation process in the [Generator repository](https://github.com/stackitcloud/stackit-sdk-generator).
```

examples/waiter/waiter.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/stackitcloud/stackit-sdk-go/core/utils"
 	"github.com/stackitcloud/stackit-sdk-go/services/dns"
+	"github.com/stackitcloud/stackit-sdk-go/services/dns/wait"
```
vicentepinto98, line 11: I would go with the alias `dnsWait` to make it more clear, since we also have the `core/wait` pkg

vicentepinto98 (approved):

examples/waiter/waiter.go
hcsa73: I'd leave it like this. It's short and clear, looking at the code you'll easily assume that it's the wait for the dns, and it's unlikely conflict with other wait packages will happen: I don't think you ever want to import `core/wait`, and if you end up using waiters from multiple services, one can then use `[SERVICE]wait` for an alias

(Also small note, package names should be lower case be convention: https://go.dev/blog/package-names, so I'd keep the alias lower case as well)

hcsa73: Had to remove waiter example, the lint was complaining that the dns release didn't have the `wait` package

vicentepinto98 (approved):

PR #107, opened 2023-10-09
Update README with key flow info

vicentepinto98:


README.md
```diff
@@ -126,32 +126,41 @@ To configure it, follow this steps:
...
-   }
-   ```
+1.  In the Portal, go to the `Service Accounts` tab, choose a `Service Account` and go to `Service Account Keys` to create a key.
+    - You can create your own RSA key-pair or have the Portal generate one for you.
+2.  Save the content of the service account key and the corresponding private key by copying them or saving them in a file.
+
+    **Hint:** if you have generated the RSA key-pair using the Portal, to save the private key in a PEM encoded file follow these steps:
+
+    - Download the service account key as a PEM file
+    - Extract the private key from the service account key using this command: `openssl storeutl -keys <path/to/sa_key_pem_file> > private.key`
```
joaopalet, line 136: We can maybe make a bit shorter with
```suggestion
    **Hint:** If you have generated the RSA key-pair using the Portal, you can save the private key in a PEM encoded file by downloading the service account key as a PEM file and using `openssl storeutl -keys <path/to/sa_key_pem_file> > private.key` to extract the private key from the service account key.
```

joaopalet (approved):

joaopalet (approved):

PR #108, opened 2023-10-09
Add option to lint and test automatically generated files only

hcsa73:


scripts/test-go.sh
```diff
@@ -15,13 +19,19 @@ else
     exit 1
 fi
 
-echo ">> Testing core"
-cd ${CORE_PATH}
-go test ./... ${GOTEST_ARGS}
+if [ ! "${SKIP_NON_GENERATED_FILES}" = true ]; then
```
joaopalet, line 22: Could we safely adjust the condition? to be `!= true` or flip it to be `= false`?

scripts/lint-golangci-lint.sh
```diff
@@ -17,20 +20,28 @@ else
...
 for service_dir in ${SERVICES_PATH}/*; do
     service=$(basename ${service_dir})
     echo ">> Linting service ${service}"
     cd ${service_dir}
-    golangci-lint run ${GOLANG_CI_ARGS}
+    if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
+        golangci-lint run ${GOLANG_CI_ARGS} --skip-dirs wait
+    else
+        golangci-lint run ${GOLANG_CI_ARGS}
+    fi
```
joaopalet, line 37: Maybe same here, although here we have the `--skip-dirs` which already gives some more context

CONTRIBUTION.md
```diff
@@ -21,8 +21,8 @@ Check the [Getting Started](README.md#getting-started) and [Authentication](READ
 These commands can be executed from the project root:
 
 - `make project-tools`: get the required dependencies
-- `make lint`: lint the code and the examples and sync dependencies
-- `make test`: run unit tests
+- `make lint`: lint the code and the examples and sync dependencies. To only lint automatically generated files, run with `skip-non-generated-files=true`
+- `make test`: run unit tests. To only test automatically generated files, run with `skip-non-generated-files=true`
```
joaopalet, line 25: Will the command look like `make lint skip-non-generated-files=true`? Or `make lint --skip-non-generated-files=true`

Makefile
```diff
@@ -11,17 +11,19 @@ project-tools:
...
-	@$(SCRIPTS_BASE)/lint-golangci-lint.sh
+	@$(SCRIPTS_BASE)/lint-golangci-lint.sh ${skip-non-generated-files}
 
 sync-tidy:
 	@echo "Syncing and tidying dependencies"
 	@$(SCRIPTS_BASE)/sync-tidy.sh
 
-lint: sync-tidy lint-golangci-lint
+lint: sync-tidy
+	@$(MAKE) --no-print-directory lint-golangci-lint skip-non-generated-files=${skip-non-generated-files}
```
vicentepinto98, line 21: what does `--no-print-directory` do?

scripts/test-go.sh
```diff
@@ -1,8 +1,12 @@
 #!/bin/bash
 # This script tests the SDK modules
+# To skip manually maintained files, pass an extra "true" argument
 # Pre-requisites: Go
 set -eo pipefail
 
+SKIP_NON_GENERATED_FILES="${1:-false}"
+echo "${SKIP_NON_GENERATED_FILES}"
```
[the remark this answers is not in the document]
hcsa73: Test leftover, removed it

scripts/test-go.sh
hcsa73: `!= true` is what is currently doing.

You can't flip it to `= false` because bash is not forcing this to be `true` or `false`. So if you were to pass `"banana"`, `"${SKIP_NON_GENERATED_FILES}" = true` is false and `"${SKIP_NON_GENERATED_FILES}" = false` is also false.

I've added input cleanup at the top to make the script clearer, PTAL

CONTRIBUTION.md
hcsa73: The first one. Added for clarity

Makefile
hcsa73: Makes it so that Make does not say `make[1]: Entering directory '/path/to/dir`

joaopalet (approved):

PR #118, opened 2023-10-23
Generator: update SDK /services

stackit-pipeline:
Generated from GitHub run 6613263619 (https://github.com/stackitcloud/stackit-sdk-generator/actions/runs/6613263619)

PR #119, opened 2023-10-23
Update dependencies

hcsa73:
After releases following #118


joaopalet (approved):
Alternatively we could let Renovate catch these, right?

hcsa73:
> Alternatively we could let Renovate catch these, right?

Yes

PR #131, opened 2023-10-26
Revamp wait package

hcsa73:
- Add generic to signature of wait function
- Rename types and fields, add comments
- Uniformize `SetXXX` signatures
- Rewrite tests


examples/waiter/waiter.go
```diff
@@ -38,20 +38,13 @@ func main() {
 	}
 
 	zoneId := *createZoneResp.Zone.Id
+
+	// The following will wait until the DNS zone is finshed being created
```
joaopalet, line 42:
```suggestion
	// The following will wait until the DNS zone creation has finished
```

core/wait/wait.go
```diff
@@ -12,72 +12,77 @@ import (
 
 var RetryHttpErrorStatusCodes = []int{http.StatusBadGateway, http.StatusGatewayTimeout}
 
-type WaitFn func() (res interface{}, done bool, err error)
-
-type Handler struct {
-	fn                WaitFn
+// AsyncActionCheck reports whether a specific async action has finished.
+//   - waitFinished == true if the async action is finished, false otherwise.
+//   - response contains data regarding the current state of the resource targeted by the async action (if applicable). resource != nil if waitFinished == true.
```
joaopalet, line 17: > resource != nil if waitFinished == true.

Did you mean `resource != nil if (...)`? Maybe we have cases where `resource` is `nil` even though the wait has finished, so we could rephrase it to `response == nil if waitFinished == false`.

core/wait/wait.go
```diff
@@ -12,72 +12,77 @@ import (
...
 }
 
-// SetThrottle sets the duration between func triggering
-func (w *Handler) SetThrottle(d time.Duration) error {
-	if d == 0 {
-		return fmt.Errorf("throttle can't be 0")
-	}
-	w.throttle = d
-	return nil
+// SetThrottle sets the duration between func triggering.
```
joaopalet, line 41: Maybe we could make this more explicit? Something like

```suggestion
// SetThrottle sets the time interval between each checkFn triggering.
```

core/wait/wait.go
```diff
@@ -12,72 +12,77 @@ import (
...
+type AsyncActionCheck[T any] func() (waitFinished bool, response *T, err error)
+
+// AsyncActionHandler handles waiting for a specific async action to be finished.
+type AsyncActionHandler[T any] struct {
+	checkFn           AsyncActionCheck[T]
 	sleepBeforeWait   time.Duration
 	throttle          time.Duration
 	timeout           time.Duration
-	tempErrRetryLimit int
+	retryLimitTempErr int
```
joaopalet, line 27: I think this read a bit better as it was before (we should also update in the `Set...` function)

core/wait/wait.go
```diff
@@ -12,72 +12,77 @@ import (
...
 
-type WaitFn func() (res interface{}, done bool, err error)
-
-type Handler struct {
-	fn                WaitFn
+// AsyncActionCheck reports whether a specific async action has finished.
+//   - waitFinished == true if the async action is finished, false otherwise.
+//   - response contains data regarding the current state of the resource targeted by the async action (if applicable). resource != nil if waitFinished == true.
+//   - err != nil if there was an error checking if the async action finished, or if it finished unsuccessfully.
+type AsyncActionCheck[T any] func() (waitFinished bool, response *T, err error)
```
joaopalet, line 19: What was the reasoning behind switching the output order?

core/wait/wait_test.go
```diff
@@ -8,66 +8,135 @@ import (
...
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	oapiError "github.com/stackitcloud/stackit-sdk-go/core/oapierror"
 )
 
+// Options used for comparing AsyncActionHandler
+var cmpOpts = []cmp.Option{
+	cmp.AllowUnexported(AsyncActionHandler[interface{}]{}),
+	cmpopts.IgnoreFields(AsyncActionHandler[interface{}]{}, "checkFn"),
```
joaopalet, line 18: Why are we ingoring the `checkFn` field?

core/wait/wait_test.go
```diff
@@ -8,66 +8,135 @@ import (
...
 			}
-			if got != nil && tt.want != nil {
-				if got.Error() != tt.want.Error() {
-					t.Errorf("Wait.SetThrottle() = %v, want %v", got, tt.want)
-				}
+			if got := w.SetThrottle(tt.args.d); !cmp.Equal(got, tt.want, cmp.AllowUnexported(AsyncActionHandler[interface{}]{})) {
+				t.Errorf("Wait.SetThrottle() = %v, want %v", got, tt.want)
 			}
 		})
 	}
```
joaopalet, line 98: Can be removed (same on below tests)

core/wait/wait_test.go
```diff
@@ -169,25 +267,239 @@ func TestSetTimeout(t *testing.T) {
...
+
+			if tt.wantErr && (err == nil) {
+				t.Errorf("expected error but got none")
+			}
+			if !tt.wantErr && (err != nil) {
+				t.Errorf("expected no error but got \"%v\"", err)
+			}
+			if (err == nil) && (resp == nil) {
+				t.Errorf("got nil err but nil resp")
+			}
```
joaopalet, line 493: Similar to a comment I made above, can't we have `checkFn`s returning nil? If the API doesn't have a response body

core/wait/wait.go
hcsa73: I did mean `resource != nil if waitFinished == true`.

It's technically possible for the function to not want to return anything when `waitFinished == true`, and the handler's `WaitWithContext` doesn't care if `resp` is `nil` or not. I'd say it's very unlikely that such an use case happens, though: for the function to decide that the wait is over, it must have done by using some API response, and so it can return that API response that proves the wait has finished (no matter if successful or unsuccessful).

I've checked, and not all usages of `AsyncActionHandler` this rule. Eg, `CreateLoadBalancerWaitHandler`, upon finding that the load balancer's creation failed, returns `true, nil, fmt.Errorf(...)`. However, it can return `true, s, fmt.Errorf(...)`, where `s` is the API response that holds the failed state of the API.

I'd suggest enforcing this rule, by 
- Checking if `resp != nil` in `WaitWithContext`
- Removing the `(if applicable)` portion of the comment
- Adding in the documentation of `WaitWithContext` that `resp != nil` if `err != nil`
- Fix usages of `AsyncActionHandler`

hcsa73: We can also enforce a weaker rule, that `resource != nil if waitFinished == true && err != nil`.

core/wait/wait.go
hcsa73: Reworded it a bit

core/wait/wait.go
hcsa73: Just that `waitFinished` is "more important", in the sense that this function gets called repeatedly until is equals `true`. Also `WaitWithContext`'s signature matches this one without the `waitFinished`, and it looked weird for it to be between the other two outputs

core/wait/wait_test.go
hcsa73: Apparently `cmp` can't compare functions

core/wait/wait_test.go
hcsa73: Ooops, leftover code

core/wait/wait_test.go
hcsa73: Check comment above

core/wait/wait_test.go
hcsa73: Added a comment for this

core/wait/wait.go
hcsa73: After discussion, will remove the rule and keep it generic

hcsa73:
Changed the rule for handlers for which the response part doesn't apply, uniformed implementation of Handlers and adapted tests

joaopalet (approved):

PR #153, opened 2023-11-10
Update core.KeyFlow struct to support CLI authentication

vicentepinto98:
- Revert Token field to private
- Add getter function for token field
- Improve comments


core/clients/key_flow.go
```diff
@@ -152,19 +160,19 @@ func (c *KeyFlow) RoundTrip(req *http.Request) (*http.Response, error) {
 	return c.doer(c.client, req, c.config.ClientRetry)
 }
 
-// GetAccessToken returns short-lived access token
+// GetAccessToken returns a short-lived access token and also stores the access and refresh tokens
```
hcsa73, line 163: Stores where?

vicentepinto98: rephrased

hcsa73 (approved):

PR #167, opened 2023-11-15
Add SetToken() to KeyFlow

vicentepinto98:
Support Authentication in the CLI:
- Reading custom token and jwks endpoints from env var is moved to the auth pkg
- Add SetToken to KeyFlow to allow setting the tokens directly

### DIFF
--- a/core/auth/auth.go
+++ b/core/auth/auth.go
@@ -150,6 +150,19 @@ func KeyAuth(cfg *config.Configuration) (http.RoundTripper, error) {
 		return nil, fmt.Errorf("configuring key authentication: private key could not be found: %w", err)
 	}
 
+	if cfg.TokenCustomUrl == "" {
+		tokenCustomUrl, tokenUrlSet := os.LookupEnv("STACKIT_TOKEN_BASEURL")
+		if tokenUrlSet {
+			cfg.TokenCustomUrl = tokenCustomUrl
+		}
+	}
+	if cfg.JWKSCustomUrl == "" {
+		jwksCustomUrl, jwksUrlSet := os.LookupEnv("STACKIT_JWKS_BASEURL")
+		if jwksUrlSet {
+			cfg.JWKSCustomUrl = jwksCustomUrl
+		}
+	}
+
 	keyCfg := clients.KeyFlowConfig{
 		ServiceAccountKey: cfg.ServiceAccountKey,
 		PrivateKey:        cfg.PrivateKey,
--- a/core/clients/key_flow.go
+++ b/core/clients/key_flow.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"strings"
 	"time"
 
@@ -25,10 +24,11 @@ const (
 	PrivateKey            = "STACKIT_PRIVATE_KEY"
 	ServiceAccountKeyPath = "STACKIT_SERVICE_ACCOUNT_KEY_PATH"
 	PrivateKeyPath        = "STACKIT_PRIVATE_KEY_PATH"
+	tokenAPI              = "https://service-account.api.stackit.cloud/token" //nolint:gosec // linter false positive
+	jwksAPI               = "https://service-account.api.stackit.cloud/.well-known/jwks.json"
+	defaultTokenType      = "Bearer"
+	defaultScope          = ""
 )
-
-var tokenAPI = "https://service-account.api.stackit.cloud/token" //nolint:gosec // linter false positive
-var jwksAPI = "https://service-account.api.stackit.cloud/.well-known/jwks.json"
 
 // KeyFlow handles auth with SA key
 type KeyFlow struct {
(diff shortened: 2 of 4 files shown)


core/clients/key_flow_test.go
```diff
@@ -113,6 +118,67 @@ func TestKeyFlowInit(t *testing.T) {
...
+			c := &KeyFlow{}
+			err = c.SetToken(accessToken, tt.refreshToken)
+
+			if (err != nil) != tt.wantErr {
+				t.Errorf("KeyFlow.SetToken() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if err == nil {
+				expectedKeyFlowToken := &TokenResponseBody{
+					AccessToken:  accessToken,
+					ExpiresIn:    int(time.Now().Add(24 * time.Hour).Unix()),
```
joaopalet, line 169: Hm, not sure how this timestamp is matching the one defined above since `time.Now` is probably different between the two calls. Probably we should define the timestamp above in a variable and use it in both places

core/clients/key_flow.go
```diff
@@ -131,6 +123,30 @@ func (c *KeyFlow) Init(cfg *KeyFlowConfig) error {
...
+		return fmt.Errorf("parse access token to read expiration time: %w", err)
+	}
+	exp, err := parsedAccessToken.Claims.GetExpirationTime()
+	if err != nil {
+		return fmt.Errorf("get expiration time from access token: %w", err)
+	}
+
+	c.token = &TokenResponseBody{
+		AccessToken:  accessToken,
+		ExpiresIn:    int(exp.Time.Unix()),
```
joaopalet, line 142: Not related to this PR, but would it make sense to change this field name to `ExpiresAt`? As it's expiration time and not a countdown to it

core/clients/key_flow_test.go
```diff
@@ -113,6 +118,67 @@ func TestKeyFlowInit(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var accessToken string
+			var err error
+			if tt.tokenInvalid {
+				accessToken = "foo"
+			} else {
+				accessTokenJWT := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.RegisteredClaims{
+					ExpiresAt: jwt.NewNumericDate(time.Now().Add(24 * time.Hour))})
```
joaopalet, line 153: And here is also called `ExpiresAt`

core/clients/key_flow.go
```diff
@@ -131,6 +123,30 @@ func (c *KeyFlow) Init(cfg *KeyFlowConfig) error {
 	return c.validate()
 }
 
+// SetToken can be used to set an access and refresh token manually in the client
+// the other fields in the token field are determined by inspecting the token or setting default values
```
joaopalet, line 127:
```suggestion
// SetToken can be used to set an access and refresh token manually in the client.
// The other fields in the token field are determined by inspecting the token or setting default values.
```

core/auth/auth.go
```diff
@@ -150,6 +150,19 @@ func KeyAuth(cfg *config.Configuration) (http.RoundTripper, error) {
 		return nil, fmt.Errorf("configuring key authentication: private key could not be found: %w", err)
 	}
 
+	if cfg.TokenCustomUrl == "" {
```
joaopalet, line 153: Don't we need to also check if `cfg.TokenCustomUrl == nil`?

core/auth/auth.go
```diff
@@ -150,6 +150,19 @@ func KeyAuth(cfg *config.Configuration) (http.RoundTripper, error) {
...
+		tokenCustomUrl, tokenUrlSet := os.LookupEnv("STACKIT_TOKEN_BASEURL")
+		if tokenUrlSet {
+			cfg.TokenCustomUrl = tokenCustomUrl
+		}
+	}
+	if cfg.JWKSCustomUrl == "" {
```
joaopalet, line 159: Same here

core/clients/key_flow_test.go
```diff
@@ -34,7 +36,10 @@ const saKeyStrPattern = `{
 	"validUntil": "2024-03-22T18:05:41Z"
 }`
 
-var saKey = fmt.Sprintf(saKeyStrPattern, uuid.New().String(), uuid.New().String(), uuid.New().String())
+var (
+	saKey          = fmt.Sprintf(saKeyStrPattern, uuid.New().String(), uuid.New().String(), uuid.New().String())
+	testSigningKey = []byte("Test")
+)
```
joaopalet, line 42:
```suggestion
const (
	saKey          = fmt.Sprintf(saKeyStrPattern, uuid.New().String(), uuid.New().String(), uuid.New().String())
	testSigningKey = []byte("Test")
)
```

core/clients/key_flow.go
vicentepinto98: The field name is `ExpiresIn` to match the field name on the response from the ServiceAccounts API `/token` endpoint. Check https://docs.api.eu01.stackit.cloud/documentation/service-account/version/v2#tag/Public/operation/post-token

vicentepinto98: I agree that it should be `ExpiresAt`, but that would have to be renamed in the API itself

core/clients/key_flow_test.go
vicentepinto98: see comment above

core/auth/auth.go
joaopalet: Ah it's a string, I was tricked by the `omitempty`. So if the field is a string and is not defined in a struct the default is `""`?

vicentepinto98: its a string

core/auth/auth.go
vicentepinto98: its a string, it cannot be nil

core/clients/key_flow_test.go
vicentepinto98: They are not constant. The first one results from string formatting and the second because it is an array (it can be appended to, for example)

core/clients/key_flow_test.go
vicentepinto98: The test was passing so I guess its fast enough that they match, but I've changed it to be more robust

core/clients/key_flow_test.go
joaopalet: Ah I though we could define them as `const`s and not allow changes

joaopalet (approved):

PR #177, opened 2023-11-16
Check for expired token in validateToken() in KeyFlow

vicentepinto98:


core/clients/key_flow_test.go
```diff
@@ -38,7 +38,8 @@ const saKeyStrPattern = `{
 
 var (
 	saKey          = fmt.Sprintf(saKeyStrPattern, uuid.New().String(), uuid.New().String(), uuid.New().String())
-	testSigningKey = []byte("Test")
+	testSigningKey = []byte(`Test`)
+	testJwks       = []byte(`{ "keys": [ { "kty":"oct", "kid":"test", "alg":"HS256" } ] }`)
```
vicentepinto98, line 42: currently this could be any value, but I will leave it like this for the future because we should improve these tests and this can be a starting point

hcsa73 (approved):

PR #203, opened 2023-12-15
Simplify key flow: extract private key from service account key if present

vicentepinto98:


core/auth/auth.go
```diff
@@ -139,15 +139,36 @@ func TokenAuth(cfg *config.Configuration) (http.RoundTripper, error) {
 
 // KeyAuth configures the key flow and returns an http.RoundTripper
 // that can be used to make authenticated requests using an access token
+// The KeyFlow requires a service account key and a private key.
```
hcsa73, line 142:
```suggestion
//
// The KeyFlow requires a service account key and a private key.
```
So the comments are shown better formatted

core/auth/auth.go
```diff
@@ -139,15 +139,36 @@ func TokenAuth(cfg *config.Configuration) (http.RoundTripper, error) {
...
+// If the private key is not provided explicitly, KeyAuth will check if there is one included
+// in the service account key and use that one.
 func KeyAuth(cfg *config.Configuration) (http.RoundTripper, error) {
 	err := getServiceAccountKey(cfg)
 	if err != nil {
 		return nil, fmt.Errorf("configuring key authentication: service account key could not be found: %w", err)
 	}
 
+	// Unmarshal service account key to check if private key is present
```
hcsa73, line 151: The `KeyAuth` comment states (implicitly, that's how I read it at least) an explicitly passed private key takes precedence over the one in the service account key. But here it's the other way around.

Not sure which one is best. Usually I'd say to use the explicit option first, but here if the private key is given in the service account key then I don't see why that wouldn't match the service account key. So I think doing it as you're doing is fine, just clarify the comment

core/auth/auth.go
```diff
@@ -139,15 +139,36 @@ func TokenAuth(cfg *config.Configuration) (http.RoundTripper, error) {
...
+
+	// Try to get private key from configuration, environment or credentials file
 	err = getPrivateKey(cfg)
 	if err != nil {
-		return nil, fmt.Errorf("configuring key authentication: private key could not be found: %w", err)
+		if extractedPrivateKey == "" {
+			return nil, fmt.Errorf("configuring key authentication: private key is not part of the service account key and could not be found: %w", err)
+		}
+		// Use the extracted private key if it could not be found elsewhere
+		cfg.PrivateKey = extractedPrivateKey
```
hcsa73, line 171: I'm confused by this logic. Maybe I'm seeing things wrong, but what you're doing here is:
- Get private key from explicit setting
- If you successfully extracted the private key from the service account key (aka, `extractedPrivateKey != ""`), replace the explicit private key with the extracted one

Doesn't that mean you never use the explicit private key? You're always throwing an error or overriding it with the extracted one

I'd expect something like
```
if extractedPrivateKey == "" {
    err = getPrivateKey(cfg)
    if err != nil {
        return nil, fmt.Errorf("configuring key authentication: private key is not part of the service account key and could not be found: %w", err)
    }
    // getPrivateKey sets cfg.PrivateKey (I assume), so nothing to do?
}
```

core/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.7.5 (2024-01-09)
+
+- **Improvement:** When using the key flow, the SDK will extract the private key from the service account key and use it, if no private key is provided in the configuration, through environment variable or in the credentials file. This makes it simpler to use the key flow: if you create a service account key including the private key, you don't need to provide the private key separately anymore
```
hcsa73, line 3: Remainder to update this if the preccedence order changes

core/clients/key_flow.go
```diff
@@ -200,19 +202,15 @@ func (c *KeyFlow) configureHTTPClient() {
 }
 
 // validate the client is configured well
-func (c *KeyFlow) validate() error {
-	if c.config.ServiceAccountKey == "" {
+func (c *KeyFlow) validate() (err error) {
```
hcsa73, line 205: Why this change? It's unneccessary, doesn't hurt anything, seems just random

core/auth/auth.go
vicentepinto98: The explicit private key takes precedence over the one in the service account key, I'll improve the comment.

core/clients/key_flow.go
vicentepinto98: It was necessary for variable assignment, but I updated it to keep the function signature

core/auth/auth.go
vicentepinto98: I have moved the extraction of the private key inside the block that is only executed if we cannot get the private key explicitly.

hcsa73 (approved):

PR #205, opened 2023-12-18
Replace k8s.io/apimachinery with cenkalti/backoff

vicentepinto98:


core/clients/clients.go
```diff
@@ -55,31 +55,34 @@ func Do(client *http.Client, req *http.Request, cfg *RetryConfig) (resp *http.Re
...
-			return true, nil
-		}).WithContext(),
-	)
+			return backoff.Permanent(err)
+		}
+		if maxRetries > 0 && resp != nil {
+			if resp.StatusCode == http.StatusBadGateway ||
+				resp.StatusCode == http.StatusGatewayTimeout {
+				maxRetries--
+				return fmt.Errorf("requested returned a gateway error")
```
hcsa73, line 81:
```suggestion
				return fmt.Errorf("request returned a gateway error")
```

hcsa73 (approved):
Looks good! Left just a minor comment

core/clients/clients.go
hcsa73: Not relevant

PR #206, opened 2023-12-19
Schema and method name improvements

hcsa73:
Standardizes names across service modules, and improves them based on user feedback


services/argus/CHANGELOG.md
```diff
@@ -1,3 +1,73 @@
+## v0.8.0 (2023-12-19)
+
+API methods and structs were renamed to have the same look and feel and according to user feedback.
```
joaopalet, line 3:
```suggestion
API methods and structs were renamed to have the same look and feel across all STACKIT services and according to user feedback.
```

CHANGELOG.md
```diff
@@ -1,3 +1,135 @@
+## Release (2023-12-19)
+
+API methods and structs were renamed to have the same look and feel and according to user feedback.
```
joaopalet, line 3:
```suggestion
API methods and structs were renamed to have the same look and feel across all services and according to user feedback.
```

joaopalet: Same for the service changelogs

CHANGELOG.md
```diff
@@ -1,3 +1,135 @@
...
+
```
joaopalet, line 4: Should we also mention here the main ideas that drove the changes? E.g. `Get<something plural>` to `List...`, `Update...` and `PartialUpdate...`

hcsa73: Added summary, please check

services/dns/wait/wait.go
```diff
@@ -46,8 +46,8 @@ func CreateZoneWaitHandler(ctx context.Context, a APIClientInterface, projectId,
 	return handler
 }
 
-// UpdateZoneWaitHandler will wait for zone update
-func UpdateZoneWaitHandler(ctx context.Context, a APIClientInterface, projectId, instanceId string) *wait.AsyncActionHandler[dns.ZoneResponse] {
+// PartialUpdateZoneWaitHandler will wait for zone update
```
joaopalet, line 49: Is is relevant for the waiter if the update was partial or not? If there is is both a `PartialUpdate...` and `Update...`, we would only need one waiter for both updates, right?

hcsa73: It's so it matches the corresponding method.

Added it to the changelogs and created a `PartialUpdateInstanceWaitHandler` for mongodbflex (which has both `Update` and `PartialUpdate`), which I think is the only case where we have waiters for update operations that can be both partial and full. Please check

joaopalet (approved):

PR #289, opened 2024-01-24
Add 2024-01-15 releases to changelogs, fix 2024-01-24 versions

hcsa73:


joaopalet (commented):
We should also add it to the root changelogs file, right?

joaopalet (approved):

PR #304, opened 2024-02-01
Update renovate gh action

vicentepinto98:


.github/workflows/renovate.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@v39.2.4
+        uses: renovatebot/github-action@v40
```
hcsa73, line 16: It doesn't require a minor and patch version? I thought it did

vicentepinto98: I dont think so, the action above also has just major version

hcsa73: That's not enough, that depends on what each GitHub action has released, and different actions have different releases
Also it doesn't work: https://github.com/stackitcloud/stackit-sdk-go/actions/runs/7744970877/job/21119926497

.github/workflows/renovate.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Self-hosted Renovate
-        uses: renovatebot/github-action@v39.2.4
+        uses: renovatebot/github-action@v40.0.0
```
hcsa73, line 16:
```suggestion
        uses: renovatebot/github-action@v40.0.2
```
https://github.com/renovatebot/github-action/releases

hcsa73 (approved):

PR #314, opened 2024-02-06
Fix postgresflex instance creation wait failing when API treturns 404 right after creation request

hcsa73:


services/postgresflex/wait/wait.go
```diff
@@ -72,7 +72,8 @@ func CreateInstanceWaitHandler(ctx context.Context, a APIClientInstanceInterface
 		}
 		return false, nil, nil
 	})
-	handler.SetTimeout(45 * time.Minute)
+	// Sleep before wait is set because sometimes API returns 404 right after creation request
+	handler.SetTimeout(45 * time.Minute).SetSleepBeforeWait(15 * time.Second)
```
joaopalet, line 76: Isn't 15 too much?

joaopalet: Although instance creations always takes much more than that so probably it's still fine

joaopalet (approved):

services/postgresflex/wait/wait.go
hcsa73: Yeah, that's the thing, it's not worth being picky on seconds wait when instance creation takes minutes

PR #324, opened 2024-02-16
Move `WithCaptureHTTPResponse`

hcsa73:
To avoid confusion due to it not being a configuration option


core/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.8.0 (2024-02-16)
+
+- **Feature:** `WithCaptureHTTPResponse` was moved from `config` package to a new `runtime` package, to avoid confusion due to it not being a configuration option
```
vicentepinto98, line 3: Add explicitly that `config.WithCaptureHTTPResponse` is deprecated and that `runtime.WithCaptureHTTPResponse` should be used instead, saying it was "moved" is not clear

core/runtime/runtime.go
```diff
@@ -0,0 +1,14 @@
...
+	"context"
+	"net/http"
+
+	"github.com/stackitcloud/stackit-sdk-go/core/config"
+)
+
+// WithCaptureHTTPResponse adds the raw HTTP response retrieval annotation to the parent context.
+// The resp parameter will contain the raw HTTP response after the request has completed.
+func WithCaptureHTTPResponse(parent context.Context, resp **http.Response) context.Context {
+	return context.WithValue(parent, config.ContextHTTPResponse, resp)
```
joaopalet, line 13: Maybe it also makes sense to move `ContextHTTPResponse` to the `runtime` package

joaopalet: Then `runtime` would not need to have a dependency on `config`

joaopalet: Does it break if we leave the current one during the deprecation period?
Then we would update the template to use the `runtime.ContextHTTPResponse` instead

joaopalet: After than I think we could safely remove the deprecated method and var in `config`

core/CHANGELOG.md
hcsa73: Lint was complaining because the latest `core` release doesn't have a `runtime` package, which the example was calling
And it also complains if it uses a deprecated function

So I'll have to do a release with the new method, and another with the old method deprecated and the example changed

Updated the changelog and added to the global changelog, please check

core/runtime/runtime.go
hcsa73: Then `config` imports `runtime` to use the new `ContextHTTPResponse`
That leaves us with `config` importing `runtime` instead of `runtime` importing `config`
Is that better?

joaopalet: What I meant is that:
- `WithCaptureHTTPResponse` in `config`  uses the current `ContextHTTPResponse` from `config`
  - And these would be removed at the same time after the deprecation period
- While the new `WithCaptureHTTPResponse` in `runtime`  uses the new `ContextHTTPResponse` from `runtime`

hcsa73: That breaks `WithCaptureHTTPResponse` from `config`

When the templates change, then they'll be checking `runtime.ContextHTTPResponse` which is different than `config.ContextHTTPResponse`. Doesn't matter that they have the same struct name and are built the same way, if one doesn't import from the other then they're different for Go.
(This is intentional, so that a package can use context keys with the same name without having to worry about collisions with keys from other packages)

joaopalet: Ah you're right, because users can update their service version and keep using the old method from config

joaopalet (approved):

PR #327, opened 2024-02-16
Remove Key Flow unused methods

hcsa73:


joaopalet (commented):
LGTM, maybe would just highlight in the changelogs that this was not used anywhere in the module

core/clients/key_flow.go
```diff
@@ -99,14 +99,6 @@ func (c *KeyFlow) GetServiceAccountEmail() string {
 	return c.key.Credentials.Iss
 }
 
-// GetToken returns the token field
```
vicentepinto98, line 102: This method is used in the CLI for the keyFlow

vicentepinto98: We can't remove it, unless we would export the `token` field. But in this case I think it's not worth it to make the breaking change. I'm not sure why I didn't just make the field public at that time (I thought I did :thinking: )

vicentepinto98 (changes requested):

vicentepinto98 (approved):

vicentepinto98 (approved):

PR #333, opened 2024-02-20
Add token refresh in background

hcsa73:


core/clients/key_flow_test.go
```diff
@@ -182,12 +182,7 @@ func TestSetToken(t *testing.T) {
 	}
 }
 
-func TestKeyFlowValidateToken(t *testing.T) {
```
hcsa73, line 185: Leftover from past PR

core/clients/key_flow.go
```diff
@@ -294,26 +327,44 @@ func (c *KeyFlow) parseTokenResponse(res *http.Response) error {
...
 	}
 	if res.StatusCode != http.StatusOK {
-		return fmt.Errorf("received: %+v", res)
+		body, err := io.ReadAll(res.Body)
+		if err != nil {
+			// Fail silently, omit body from error
+			// We're trying to show error details, so it's unnecessary to fail because of this err
+			body = []byte{}
+		}
+		return &oapierror.GenericOpenAPIError{
```
hcsa73, line 336: Changed to be an `oapierror.GenericOpenAPIError` so the token refresh can fetch the status code

core/config/config.go
```diff
@@ -85,6 +85,12 @@ type Configuration struct {
 	HTTPClient            *http.Client
 	RetryOptions          *clients.RetryConfig
 
+	// If != nil, a goroutine will be launched that will refresh the service account's access token when it's close to being expired.
+	// The goroutine is killed whenever this context is cancelled.
+	//
+	// Only has effect for key flow
+	TokenRefreshInBackgroundContext context.Context
```
vicentepinto98, line 92: Although I understand it, the name is not very user-friendly. I wonder if context is even necessary
Maybe `BackgroundTokenRefresh` or  `BackgroundTokenRefreshContext` ?

core/clients/key_flow.go
```diff
@@ -97,13 +104,19 @@ func (c *KeyFlow) GetServiceAccountEmail() string {
 
 // GetToken returns the token field
 func (c *KeyFlow) GetToken() TokenResponseBody {
+	c.tokenMutex.RLock()
+	defer c.tokenMutex.RUnlock()
+
 	if c.token == nil {
 		return TokenResponseBody{}
 	}
+	// Returned struct is passed by value (because it's a struct: https://go.dev/play/p/0VQGLfP6vsg)
```
vicentepinto98, line 113: interesting use of Go playground in a comment :laughing: 

do we really need it? maybe the just the rest of the comment is enough, to avoid noise in the code

core/clients/key_flow.go
```diff
@@ -97,13 +104,19 @@ func (c *KeyFlow) GetServiceAccountEmail() string {
 
 // GetToken returns the token field
 func (c *KeyFlow) GetToken() TokenResponseBody {
+	c.tokenMutex.RLock()
```
vicentepinto98, line 107: why do we need `RLock()` here and `Lock()` below? What's the difference?

core/clients/key_flow_token_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+	// Token refresh tries start at [Access token expiration timestamp] - [This duration]
+	timeStartBeforeTokenExpiration time.Duration
+	timeBetweenContextCheck        time.Duration
+	timeBetweenTries               time.Duration
+}
+
+// Continuously refreshes the token of a key flow, retrying if the token API returns 5xx errrors. Always returns with a non-nil error.
+//
+// To terminate this routine, close the context in refresher.keyFlow.config.TokenRefreshInBackgroundContext.
+func (refresher *tokenRefresher) refreshToken() error {
```
vicentepinto98, line 45: maybe just call this `refresh`

there's already too many `refresh` and `token` going around, plus having two methods with the same name is strange

core/clients/key_flow_token_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+// To terminate this routine, close the context in keyFlow.config.TokenRefreshInBackgroundContext.
+func refreshToken(keyflow *KeyFlow) {
+	refresher := &tokenRefresher{
+		keyFlow:                        keyflow,
+		timeStartBeforeTokenExpiration: defaultTimeStartBeforeTokenExpiration,
+		timeBetweenContextCheck:        defaultTimeBetweenContextCheck,
+		timeBetweenTries:               defaultTimeBetweenTries,
+	}
+	err := refresher.refreshToken()
+	fmt.Fprintf(os.Stderr, "Token refreshing terminated: %v", err)
```
vicentepinto98, line 31: our SDK doesn't print anything if Im not mistaken, this would be the first thing
I think it would be better to have this method return error, even if we don't do anything with it for now. Just thinking that maybe in the future there could be some extension where this would go into logs

core/clients/key_flow_token_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	startRefreshTimestamp := expirationTimestamp.Add(-refresher.timeStartBeforeTokenExpiration)
+
+	for {
+		err = refresher.waitUntilTimestamp(startRefreshTimestamp)
+		if err != nil {
+			return err
+		}
+
+		err := refresher.keyFlow.config.TokenRefreshInBackgroundContext.Err()
```
vicentepinto98, line 58: just to confirm, cancelling the context will make this return error?

core/clients/key_flow_token_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+	expirationTimestampNumeric, err := tokenParsed.Claims.GetExpirationTime()
+	if err != nil {
+		return nil, fmt.Errorf("get expiration timestamp: %w", err)
+	}
+	return &expirationTimestampNumeric.Time, nil
+}
+
+func (refresher *tokenRefresher) waitUntilTimestamp(timestamp time.Time) error {
+	for time.Now().Before(timestamp) {
```
vicentepinto98, line 97: I didnt know this worked :open_mouth:

core/clients/key_flow_token_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+		time.Sleep(refresher.timeBetweenContextCheck)
+	}
+	return nil
+}
+
+// Returns:
+//   - (true, nil) if successful.
+//   - (false, nil) if not successful but should be retried.
+//   - (_, err) if not successful and shouldn't be retried.
+func (refresher *tokenRefresher) refreshTokens() (bool, error) {
```
vicentepinto98, line 111: I would also rename this one

core/clients/key_flow.go
hcsa73: Well if you believe me, I guess not :wink:
Removed

core/clients/key_flow.go
hcsa73: Not sure where "below" is, but:
- `RLock()` locks for writes
- `Lock()` locks for reads

core/clients/key_flow_token_refresh.go
hcsa73: `refresh` really doesn't say what this does
Renamed to `continuousTokenRefresh` and renamed the filed
If you have better names, please let me know

core/clients/key_flow_token_refresh.go
hcsa73: Yes, this would be the first thing to be printed

Making this return an error doesn't help. This is called on a background async routine, that can happen at any time, and the routine just terminates when this happens.

The only options I can think of are to do nothing, print to console or add somewhere in the context. The third is useless I think (no one is going to check it, there's no obvious time to when to check it). And printing is better than having it not do anything, but we can suppress this print

core/clients/key_flow_token_refresh.go
hcsa73: Yes, this is the idiomatic way of checking the context was canceled

core/clients/key_flow.go
```diff
@@ -233,7 +263,7 @@ func (c *KeyFlow) createAccessToken() (err error) {
 	defer func() {
 		tempErr := res.Body.Close()
 		if tempErr != nil {
-			err = fmt.Errorf("closing request access token response: %w", tempErr)
+			err = fmt.Errorf("close request access token response: %w", tempErr)
 		}
 	}()
```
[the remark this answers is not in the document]
hcsa73: Good point, changed it to only override `err` when `err != nil`

core/config/config.go
vicentepinto98: I meant having `Context` in the name. But yeah let's keep it for clarity

core/clients/key_flow_continuous_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+
+var (
+	defaultTimeStartBeforeTokenExpiration = 30 * time.Minute
+	defaultTimeBetweenContextCheck        = time.Second
+	defaultTimeBetweenTries               = 5 * time.Minute
+)
+
+// Continuously refreshes the token of a key flow, retrying if the token API returns 5xx errrors. Writes to stderr when it terminates.
+//
+// To terminate this routine, close the context in keyFlow.config.TokenRefreshInBackgroundContext.
```
vicentepinto98, line 22: the comment still has the old field name

hcsa73: Fixed

core/clients/key_flow_continuous_refresh_test.go
```diff
@@ -0,0 +1,354 @@
...
+// Tests if
+//   - continuousRefreshToken() changes the token
+//   - The access token can be accessed while continuousRefreshToken() is trying to update it
+func TestContinuousRefreshTokenConcurrency(t *testing.T) {
+	// The times here are in the order of miliseconds (so they run faster)
+	// For this to work, we need to increase precision of the expiration timestamps
+	jwt.TimePrecision = time.Millisecond
+
+	// Test plan:
+	// 1) refreshToken() will trigger a token update. It will be blocked in the mockDo() routine
```
vicentepinto98, line 166: why do we need to artificially block the token update? Wouldn't it be possible to write the test in a way that the refreshToken() routine is waiting for the refreshTimestamp "naturally" while the first request is done?

Maybe Im missing something

core/clients/key_flow_token_refresh.go
vicentepinto98: I think now that you changed the others, this name actually makes more sense. This is where you actually refresh the tokens

core/clients/key_flow_continuous_refresh_test.go
hcsa73: It's intended, to test that the access token is still usable even when `continuousRefreshToken()` is trying to update the token

core/clients/key_flow_continuous_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+//   - (true, nil) if successful.
+//   - (false, nil) if not successful but should be retried.
+//   - (_, err) if not successful and shouldn't be retried.
+func (refresher *continuousTokenRefresher) refreshToken() (bool, error) {
+	err := refresher.keyFlow.createAccessTokenWithRefreshToken()
+	if err == nil {
+		return true, nil
+	}
+
+	// Should be retired if this is an API error with status code non-5xx
```
joaopalet, line 117: Did you mean `Shouldn't be retried`? Or did you actually want to use `retired` as in "retire" the goroutine?

core/clients/key_flow_continuous_refresh.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			return fmt.Errorf("check context: %w", err)
+		}
+
+		ok, err := refresher.refreshToken()
+		if err != nil {
+			return fmt.Errorf("refresh tokens: %w", err)
+		}
+		if !ok {
+			startRefreshTimestamp = startRefreshTimestamp.Add(refresher.timeBetweenTries)
```
joaopalet, line 68: What will happen if the `startRefreshTimestamp` eventually surpasses the `expirationTimestamp`? Will the API return a 5xx error and the goroutine is terminated? Or should we check this case so it doesn't run indefinetely?

core/clients/key_flow_continuous_refresh.go
joaopalet: But it says `non-5xx`, the idea is to only retry if error code is `5xx`, right?

hcsa73: Ah oops, double negative

joaopalet (approved):

PR #335, opened 2024-02-22
Automatic tag helper script

vicentepinto98:
Usage: 
- `go run automatic_tag.go --update-type [minor|patch] --ssh-private-key-file-path path --password pwd --target [core|all-services]
- `make sdk-tag-services update-type=<minor|path> ssh-private-key-file-path=<path_to_your_gh_ssh_private_key>` password=<pwd>
- `make sdk-tag-core update-type=<minor|path> ssh-private-key-file-path=<path_to_your_gh_ssh_private_key>` password=<pwd>


.github/workflows/ci.yaml
```diff
@@ -25,6 +25,15 @@ jobs:
         run: |
           make lint
           scripts/check-sync-tidy.sh
-
+      - name: Lint scripts
```
vicentepinto98, line 28: the script uses library versions that are only available in go1.21, anyway that is the only Go version where its worth testing it because we will execute it from local machine

Makefile
```diff
@@ -25,5 +31,16 @@ test-go:
...
 
+test-scripts:
+	@echo "Running Go tests for scripts"
+	@go test $(ROOT_DIR)/scripts/... ${GOTEST_ARGS}
+
 test:
 	@$(MAKE) --no-print-directory test-go skip-non-generated-files=${skip-non-generated-files}
+
+# AUTOMATIC TAG
+sdk-tag:
```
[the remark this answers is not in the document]
joaopalet: Or even `new-sdk-tags`, but probably the first one is enough

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+func automaticTagUpdate(updateType, sshPrivateKeyFilePath, password string, coreOnly bool) error {
+	tempDir, err := os.MkdirTemp("", "")
+	if err != nil {
+		return fmt.Errorf("create temporary directory: %w", err)
+	}
+
+	defer func() {
+		tempErr := os.RemoveAll(tempDir)
+		if tempErr != nil && err == nil {
+			err = fmt.Errorf("temporary directory %s could not be removed: %w", tempDir, tempErr)
```
joaopalet, line 88: I think for this to work you would need to have a named return variable:
```go
func automaticTagUpdate(...) (err error)
```

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/go-git/go-git/v5/plumbing/transport/ssh"
+	"golang.org/x/mod/semver"
+)
+
+const (
+	sdkRepo      = "git@github.com:stackitcloud/stackit-sdk-go.git"
+	patch        = "patch"
+	minor        = "minor"
+	coreOnlyFlag = "--core-only"
```
joaopalet, line 21: The name `core-only` indicates that if this flag is not passed it will create tags for all services *and* core, but that's not the case, right?

Maybe we could have instead a required argument `target` that would wither be `core` or `all-services`, as the use case is generally creating a new tag for core *or* for services. This would also be easier to extend in case we want to allow creation of tags for just a specific service

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	err = pushTags(r, publicKeys)
+	if err != nil {
+		return fmt.Errorf("push tags: %w", err)
+	}
+	return nil
+}
+
+func storeLatestTag(t *plumbing.Reference, latestTags map[string]string, coreOnly bool) map[string]string {
```
joaopalet, line 142: I think would be good to have some short descriptions of what these functions do and what is the use of each argument

joaopalet: As it might not be obvious just from the name and also because some types of the `go-git` package (e.g. `plumbing.Reference`) are a bit obscure

scripts/automatic_tag.go
joaopalet: This is absracted by the make command, but still I think it could be a nice improvement

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+		if len(splitTag) != 2 || splitTag[0] != "core" {
+			return latestTags
+		}
+
+		version := splitTag[1]
+		if semver.Prerelease(version) != "" {
+			return latestTags
+		}
+
+		// invalid (or empty) semantic version are considered less than a valid one
```
joaopalet, line 156: 👍

scripts/automatic_tag.go
joaopalet: Because this would just set the variable `err` defined in the first line of the function, but it is never returned

joaopalet: Alternatively we could just print a message

scripts/automatic_tag_test.go
```diff
@@ -0,0 +1,199 @@
...
+		t.Run(test.desc, func(t *testing.T) {
+			updatedLatestTags := storeLatestTag(test.tag, test.latestTags, test.coreOnly)
+			if !reflect.DeepEqual(test.expectedLatestTags, updatedLatestTags) {
+				t.Fatalf("Updated tags are wrong, expected %v, got %v", test.expectedLatestTags, updatedLatestTags)
+			}
+		})
+	}
+}
+
+func TestComputeUpdatedTag(t *testing.T) {
```
joaopalet, line 123: Also, could you add some test cases where the update is `minor` but the current patch version is not `0`?

E.g. `v0.1.2` -> `0.2.0`

.github/workflows/ci.yaml
```diff
@@ -25,6 +25,15 @@ jobs:
...
+        if: ${{ matrix.go-version == '1.21' }}
+        run: |
+          make lint-scripts
       - name: Test
-        run: make test
+        if: ${{ matrix.go-version != '1.18' }}
+        run: |
+          make test
```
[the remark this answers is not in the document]
vicentepinto98: > the script uses library versions that are only available in go1.21, anyway that is the only Go version where its worth testing it because we will execute it from local machine

See comment above

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+const (
+	sdkRepo      = "git@github.com:stackitcloud/stackit-sdk-go.git"
+	patch        = "patch"
+	minor        = "minor"
+	coreOnlyFlag = "--core-only"
+)
+
+var (
+	updateTypes = []string{minor, patch}
+	usage       = fmt.Sprintf("go run automatic_tag.go [minor|patch] ssh-private-key-file-path password %s", coreOnlyFlag)
```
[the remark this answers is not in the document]
vicentepinto98: I think that doesn't make a lot of sense because the script can also be run directly and it seems strange that it would give the usage of the make command

Wrong usage of the make command is not part of the scope of the script

scripts/automatic_tag_test.go
vicentepinto98: done

scripts/automatic_tag.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+func main() {
+	// Parse arguments
```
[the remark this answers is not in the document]
vicentepinto98: added it

scripts/automatic_tag.go
vicentepinto98: good suggestion, done

scripts/automatic_tag.go
```diff
@@ -0,0 +1,274 @@
...
+	updatedVersion := strings.Join(splitVersion, ".")
+
+	switch target {
+	case core:
+		if service != "core" {
+			return "", fmt.Errorf("%s target was provided but store latest tag from another service: %s", target, service)
+		}
+		return fmt.Sprintf("core/%s", updatedVersion), nil
+	case allServices:
+		return fmt.Sprintf("services/%s/%s", service, updatedVersion), nil
```
joaopalet, line 239: With this current logic, aren't we creating a `service/core/vX.Y.Z` tag when the target is `allServices`?

joaopalet: Probably `computeUpdatedTag` doesn't need to know about `target`, and this logic could be moved to `automaticTagUpdate`.

(remembering comment above regarding `service` -> `module`) This would leave this function signature as `func computeUpdatedTag(module, version, updateType string)`, just focusing on calculating the next tag for module `module` at version `version` given the `updateType`

vicentepinto98: I agree, that makes sense

vicentepinto98: > With this current logic, aren't we creating a service/core/vX.Y.Z tag when the target is allServices?
No, because `lastestTags` wouldn't have the `core` module

scripts/automatic_tag.go
```diff
@@ -0,0 +1,274 @@
...
+		if err != nil {
+			return fmt.Errorf("store latest tag: %w", err)
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("iterate over existing tags: %w", err)
+	}
+
+	for service, version := range latestTags {
```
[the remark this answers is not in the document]
vicentepinto98: done

scripts/automatic_tag.go
vicentepinto98: Done

joaopalet (approved):

PR #354, opened 2024-03-18
Extend Load Balancer creation waiter to look for errors in `Errors` field

joaopalet:


services/loadbalancer/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.9.3 (2024-03-19)
+
+- Update Load Balancer creation waiter to check for errors the instance `Errors` field and, if some are found, return even if the status is still `STATUS_PENDING`
```
vicentepinto98, line 3:
```suggestion
- Improve error handling in Load Balancer creation waiter, fixing timeout for `STATUS_PENDING` state with errors. If an error is found in the `Errors` field, it returns `true` and an error. 
```

services/loadbalancer/wait/wait.go
```diff
@@ -46,6 +47,15 @@ func CreateLoadBalancerWaitHandler(ctx context.Context, a APIClientInterface, pr
 		if s == nil || s.Name == nil || *s.Name != instanceName || s.Status == nil {
 			return false, nil, nil
 		}
+
+		if s.Errors != nil && len(*s.Errors) != 0 {
```
vicentepinto98, line 51: Could it happen that the instance has STATUS_READY and errors in the `Errors` field? In that case the waiter will fail.

We should clarify with the team if that can happen and what should we do in that case.

services/loadbalancer/wait/wait.go
```diff
@@ -46,6 +47,15 @@ func CreateLoadBalancerWaitHandler(ctx context.Context, a APIClientInterface, pr
...
+			errors := make([]string, len(*s.Errors))
+			for _, err := range *s.Errors {
+				errors = append(errors, fmt.Sprintf("%s: %s", *err.Type, *err.Description))
+			}
+			return true, s, fmt.Errorf("create failed for instance with name %s, got errors: %s", instanceName, strings.Join(errors, ";"))
```
vicentepinto98, line 56: I would also add the status to the error message

services/loadbalancer/wait/wait.go
joaopalet: Discussed it with colleague from the Load Balancer team. They think there is also an issue on their side and will investigate why errors might be set and not be reflected on the state. Nevertheless, he said we could use this errors list to determine if there is an error or not

vicentepinto98 (approved):

PR #356, opened 2024-03-19
Update ConfigureRegion for new global OAS server format

vicentepinto98:
With the migration to the stackit-api repo, the generated OAS for global APIs will have a different server url that follows this pattern:
```
"servers": [
    {
      "url": "https://identifier.api.stackit.cloud",
      "variables": {
        "region": {
          "default": "global"
        }
      }
    }
  ]
  ```
  
  This PR updates the ConfigureRegion method to take into account this change

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -13,6 +13,10 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/core/clients"
 )
 
+const (
+	global = "global"
+)
+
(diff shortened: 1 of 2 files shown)


joaopalet (approved):
So for regional APIs the URL will still contain the template, right?

vicentepinto98:
> So for regional APIs the URL will still contain the template, right?

Yes, that way we can configure multiple regions, as long as they are defined in the OAS

PR #367, opened 2024-03-28
Add waiters for SKE credentials rotation

DiogoFerrao:


services/ske/wait/wait.go
```diff
@@ -148,3 +153,59 @@ func RotateCredentialsWaitHandler(ctx context.Context, a APIClientClusterInterfa
...
+			return true, s, nil
+		}
+
+		if state == CredentialsStatePreparing {
+			return false, nil, nil
+		}
+
+		if state == CredentialsStateNever {
+			return true, s, fmt.Errorf("starting credentials rotation failed")
+		}
```
joaopalet, line 176: Why do we need a special case for the `NEVER`? According to the API documentation: `NEVER indicates that no credentials rotation has been performed using the new credentials rotation endpoints yet`, so I think we could group it into the case below where we provide the state in the error message

services/ske/wait/wait.go
```diff
@@ -148,3 +153,59 @@ func RotateCredentialsWaitHandler(ctx context.Context, a APIClientClusterInterfa
...
+
+		return true, s, fmt.Errorf("unexpected state %s while waiting for cluster credentials rotation phase", state)
```
joaopalet, line 178:
```suggestion
		return true, s, fmt.Errorf("unexpected status %s while waiting for cluster credentials rotation to be prepared", state)
```

services/ske/wait/wait.go
```diff
@@ -148,3 +153,59 @@ func RotateCredentialsWaitHandler(ctx context.Context, a APIClientClusterInterfa
...
+
+		if state == CredentialsStateCompleting {
+			return false, nil, nil
+		}
+
+		if state == CredentialsStateNever {
+			return true, s, fmt.Errorf("completing credentials rotation failed")
+		}
+
+		return true, s, fmt.Errorf("unexpected state %s while waiting for cluster credentials rotation phase", state)
```
joaopalet, line 206:
```suggestion
		return true, s, fmt.Errorf("unexpected status %s while waiting for cluster credentials rotation to be completed", state)
```

services/ske/wait/wait.go
```diff
@@ -148,3 +153,59 @@ func RotateCredentialsWaitHandler(ctx context.Context, a APIClientClusterInterfa
...
+			return true, s, nil
+		}
+
+		if state == CredentialsStateCompleting {
+			return false, nil, nil
+		}
+
+		if state == CredentialsStateNever {
+			return true, s, fmt.Errorf("completing credentials rotation failed")
+		}
```
joaopalet, line 204: Same as above

services/ske/wait/wait.go
```diff
@@ -19,6 +19,11 @@ const (
 	StateCreated                  = "STATE_CREATED"
 	StateUnhealthy                = "STATE_UNHEALTHY"
 	StateReconciling              = "STATE_RECONCILING"
+	CredentialsStateNever         = "NEVER"
```
joaopalet, line 22: Could be changed to `CredentialsRotationState...`

services/ske/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## v0.11.0 (2024-03-28)
+
+- Waiters for async operation `StartCredentialsRotationWaitHandler` and `CompleteCredentialsRotationWaitHandler`
```
joaopalet, line 3:
```suggestion
- **Feature**: Waiters for async operation `StartCredentialsRotationWaitHandler` and `CompleteCredentialsRotationWaitHandler`
```

joaopalet: Also not sure we need the link to the example, as not all new features will be included in examples

joaopalet (approved):

PR #372, opened 2024-04-03
PostgreSQL Flex: add waiter for ForceDeleteInstance and fix DeleteInstance waiter

DiogoFerrao:


services/postgresflex/wait/wait.go
```diff
@@ -106,6 +107,27 @@ func PartialUpdateInstanceWaitHandler(ctx context.Context, a APIClientInstanceIn
...
+		}
+		if s == nil || s.Item == nil || s.Item.Id == nil || *s.Item.Id != instanceId || s.Item.Status == nil {
+			return false, nil, nil
+		}
+		switch *s.Item.Status {
+		default:
+			return true, nil, fmt.Errorf("instance with id %s has unexpected status %s", instanceId, *s.Item.Status)
+		case InstanceStateDeleted:
+			return true, nil, nil
+		}
```
joaopalet, line 123: Just to clarify, then there isn't any state between `Ready` and `Deleted`? Like a `Deleting` or something

joaopalet (approved):

services/postgresflex/wait/wait_test.go
```diff
@@ -300,6 +299,53 @@ func TestDeleteInstanceWaitHandler(t *testing.T) {
...
+	tests := []struct {
+		desc             string
+		instanceGetFails bool
+		instanceState    string
+		wantErr          bool
+	}{
+		{
+			desc:             "delete_succeeded",
+			instanceGetFails: false,
+			instanceState:    InstanceStateSuccess,
```
joaopalet, line 312: Noticed now, shouldn't this be?

```suggestion
			instanceState:    InstanceStateDeleted,
```

joaopalet (approved):

services/postgresflex/wait/wait.go
DiogoFerrao: This is pending clarification from the service team, we will merge in the mean time and do a pre-release. Then release when confirmation comes.

PR #387, opened 2024-04-09
Add WithCaptureHTTPRequest

DiogoFerrao:


core/runtime/runtime.go
```diff
@@ -12,3 +12,9 @@ import (
 func WithCaptureHTTPResponse(parent context.Context, resp **http.Response) context.Context {
 	return context.WithValue(parent, config.ContextHTTPResponse, resp)
 }
+
+// WithCaptureHTTPRequest adds the raw HTTP request to the parent context.
+// The req parameter will contain the raw HTTP request.
```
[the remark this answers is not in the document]
joaopalet: `HTTP request retrieval annotation` because `WithCaptureHTTPRequest` doesn't add the raw request to the context by itself, it just adds that "annotation" to the context to signal that it should be captured when the request is executed

core/runtime/runtime.go
```diff
@@ -12,3 +12,9 @@ import (
 func WithCaptureHTTPResponse(parent context.Context, resp **http.Response) context.Context {
 	return context.WithValue(parent, config.ContextHTTPResponse, resp)
 }
+
+// WithCaptureHTTPRequest adds the raw HTTP request retrieval annotation to the parent context.
+// The req parameter will contain the raw HTTP request made to the API.
```
joaopalet, line 17:
```suggestion
// After the request has completed, the req parameter will contain the raw HTTP request made to the API.
```

joaopalet: I would still highlight the `req` variable will only be populated once the request is executed

PR #395, opened 2024-04-10
Generator: update SDK /services

stackit-pipeline:
Generated from GitHub run 8634606605 (https://github.com/stackitcloud/stackit-sdk-generator/actions/runs/8634606605)


services/argus/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.9.7 (2024-04-11)
+
+- Add Middleware logic to the client.
```
vicentepinto98, line 3: Make it more explicit:
"Support `WithMiddleware` configuration option in the client"

services/ske/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.11.3 (2024-04-11)
+
+- Add Middleware logic to the client.
+
 ## v0.11.2 (2024-04-09)
```
vicentepinto98, line 5: none of the releases that are part of this changelog were made on 9/04, so this change can just be a part of that release and there is no need to increment the version

services/argus/client.go
```diff
@@ -80,7 +80,12 @@ func NewAPIClient(opts ...config.ConfigurationOption) (*APIClient, error) {
 		return nil, fmt.Errorf("setting up authentication: %w", err)
 	}
 
-	cfg.HTTPClient.Transport = authRoundTripper
+	roundTripper := authRoundTripper
+	if cfg.Middleware != nil {
+		roundTripper = config.ChainMiddleware(roundTripper, cfg.Middleware...)
```
joaopalet, line 85: Is this method already released in a new version of `core`?

joaopalet: And the `core` dependency updated for all these services?

joaopalet (changes requested):

services/argus/client.go
vicentepinto98: I assume the lint is not failing because it uses the local version of the core? We should actually fix this to have some way that in the pipeline it would use the version that is imported, otherwise we could easily be merging this without noticing and it would be a runtime error

services/postgresql/CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 >
 > Use `github.com/stackitcloud/stackit-sdk-go/services/postgresflex` instead. For more details, check https://docs.stackit.cloud/stackit/en/bring-your-data-to-stackit-postgresql-flex-138347648.html.
 
-## v0.13.1 (2024-04-09)
+## v0.14.0 (2024-04-11)
```
vicentepinto98, line 5: Remove this entry from the changelog. Since PG is deprecated, no further releases should be done

services/argus/client.go
DiogoFerrao: Core is now updated and go.mods are using the latest version

services/loadbalancer/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.10.2 (2024-04-11)
+
+- Update `core` to [`v0.12.0`](../../core/CHANGELOG.md#v0120-2024-04-11)
```
vicentepinto98, line 3: Isn't it missing here two entrys?
```
- Set config.ContextHTTPRequest in Execute method
- Support WithMiddleware configuration option in the client
```

vicentepinto98: And then it should be minor

DiogoFerrao: This was discussed with @joaopalet. The changes in the CHANGELOG were not really applied since I reverted the entire api_default.go changes due to the missing beautification on their end. 
When João merges these changes he will update the versions and changelogs as well as the code. The changes to the client.go should also be reverted then, since without setting the cfg.Middleware, they do nothing.
But what do you think?

vicentepinto98 (approved):

PR #400, opened 2024-04-11
Add usage examples for withMiddleware and withCaptureHTTPRequest

DiogoFerrao:


examples/middleware/middleware.go
```diff
@@ -0,0 +1,63 @@
+package main
```
vicentepinto98, line 1: :+1: really cool!

examples/middleware/go.mod
```diff
@@ -0,0 +1,13 @@
+module github.com/stackitcloud/stackit-sdk-go/examples/middleware
+
+go 1.18
+
+require (
+	github.com/stackitcloud/stackit-sdk-go/core v0.10.1
```
vicentepinto98, line 6: upgrade core to the latest version

examples/middleware/middleware.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (rt roundTripperWithCapture) RoundTrip(req *http.Request) (*http.Response, error) {
+	returnStr := fmt.Sprintf("%s %s", req.Method, req.URL.String())
+
+	fmt.Println("Captured request:", returnStr)
+	fmt.Println("Status:", rt.status)
+
+	// Proceed with the original round trip
+	// This step is important to preserve the original behavior of the client
+	return rt.transport.RoundTrip(req)
```
vicentepinto98, line 34: Even cooler would be something like:

```
func (rt roundTripperWithCapture) RoundTrip(req *http.Request) (*http.Response, error) {
	returnStr := fmt.Sprintf("%s %s", req.Method, req.URL.String())

	fmt.Println("Captured request:", returnStr)

        // Proceed with the original round trip
	// This step is important to preserve the original behavior of the client
	resp, err := rt.transport.RoundTrip(req)
	fmt.Println("Status:", resp.Status)

	return resp, err
}
```

In this case it would only make sense to add one middleware though :thinking:
And I like your example because it allows to check the order of execution

vicentepinto98 (approved):

vicentepinto98 (changes requested):
core update

examples/middleware/middleware.go
DiogoFerrao: Something like what I just commited? It shows the functionality of capturing the response and maintains the demonstration of the order of execution

vicentepinto98 (approved):

examples/middleware/middleware.go
vicentepinto98: yes, now it's a really complete example :wink:

vicentepinto98 (approved):

PR #460, opened 2024-05-06
Refactor Repeated Code in getPrivateKey and getServiceAccountKey

kemalcanbora:
getPrivateKey and getServiceAccountKey functions consist of repetitive codes, to prevent this, it is sufficient to create a single function.

### DIFF
--- a/core/auth/auth.go
+++ b/core/auth/auth.go
@@ -264,67 +264,40 @@ func getServiceAccountEmail(cfg *config.Configuration) string {
 	return email
 }
 
-// getPrivateKey searches for the private key in the following order: client configuration, environment variable, credentials file.
-func getPrivateKey(cfg *config.Configuration) (err error) {
-	// get key from configuration
-	if cfg.PrivateKey == "" {
-		// get key path from configuration
-		if cfg.PrivateKeyPath == "" {
-			// get key path from environment
-			privateKeyPath, privateKeyPathSet := os.LookupEnv("STACKIT_PRIVATE_KEY_PATH")
-			if !privateKeyPathSet || privateKeyPath == "" {
-				// get key path from the credentials file
-				credentials, err := readCredentialsFile(cfg.CredentialsFilePath)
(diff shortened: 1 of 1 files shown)


core/auth/auth.go
```diff
@@ -264,67 +264,40 @@ func getServiceAccountEmail(cfg *config.Configuration) string {
...
-	// get key from configuration
-	if cfg.PrivateKey == "" {
-		// get key path from configuration
-		if cfg.PrivateKeyPath == "" {
-			// get key path from environment
-			privateKeyPath, privateKeyPathSet := os.LookupEnv("STACKIT_PRIVATE_KEY_PATH")
-			if !privateKeyPathSet || privateKeyPath == "" {
-				// get key path from the credentials file
-				credentials, err := readCredentialsFile(cfg.CredentialsFilePath)
+func getKey(cfgKey, cfgPath *string, envVar, credType credentialType, cfgFilePath string) error {
```
vicentepinto98, line 267: can you re-add the comment on top?

vicentepinto98, line 267:
```suggestion
func getKey(cfgKey, cfgKeyPath *string, envVar, credType credentialType, cfgCredFilePath string) error {
```

vicentepinto98 (commented):
Since you are refactoring this code block, it would be nice to take the chance to increase the readbility by flipping initial condition at the start.

Like this:
```
if *cfgKey != "" {return nil}
if *cfgPath == "" {
     ....
}
....
return nil
```

core/auth/auth.go
kemalcanbora: done

kemalcanbora: do you have any comment suggestion for this function because we'll remove **getServiceAccountKey** & **getPrivateKey** if you want I can add like this comment 

> getKey searches for the service account key in the following order: client configuration, environment variable, credentials file.

kemalcanbora: I did not commit to repo yet 😄

vicentepinto98 (approved):

vicentepinto98:
Thank you for your contribution! :smile:

PR #495, opened 2024-05-14
Add waiters and example for IaaS network areas

joaopalet:


services/iaas/wait/wait.go
```diff
@@ -0,0 +1,152 @@
...
+		if err != nil {
+			return false, area, err
+		}
+		if area.AreaId == nil || area.State == nil {
+			return false, area, fmt.Errorf("create failed for network area with id %s, the response is not valid: the id or the state are missing", areaId)
+		}
+		if *area.AreaId == areaId && *area.State == CreateSuccess {
+			return true, area, nil
+		}
+		return false, area, nil
```
DiogoFerrao, line 38: Typically, we return nil in the response if the waiter isn't done yet. Is there something unique going on here?

services/iaas/wait/wait.go
```diff
@@ -0,0 +1,152 @@
...
+			return false, nil, nil
+		}
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
+		if !ok {
+			return false, area, fmt.Errorf("could not convert error to oapierror.GenericOpenAPIError: %w", err)
+		}
+		if oapiErr.StatusCode != http.StatusNotFound {
+			return false, area, err
+		}
+		return true, nil, nil
```
DiogoFerrao, line 79:
```suggestion
		return true, area, nil
```

services/iaas/wait/wait.go
DiogoFerrao: I think this is valid for most `return false,...` statements in this file

joaopalet: It was intentional, I think there's no arm in also returning the current state of the object here. What we have documented for this field is `response contains data regarding the current state of the resource targeted by the async action, if applicable`

DiogoFerrao: Ok, good to know 😁

DiogoFerrao (approved):

PR #537, opened 2024-05-24
Add waiters and example for instance operations in `sqlserverflex` module

joaopalet:


services/mongodbflex/wait/wait_test.go
```diff
@@ -126,7 +126,7 @@ func TestCreateInstanceWaitHandler(t *testing.T) {
 
 			handler := CreateInstanceWaitHandler(context.Background(), apiClient, "", instanceId)
 
-			gotRes, err := handler.SetTimeout(10 * time.Millisecond).WaitWithContext(context.Background())
+			gotRes, err := handler.SetTimeout(10 * time.Millisecond).SetSleepBeforeWait(1 * time.Millisecond).WaitWithContext(context.Background())
```
DiogoFerrao, line 129: Great catch 👍

examples/sqlserverflex/sqlserverflex.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/stackitcloud/stackit-sdk-go/core/config"
+	"github.com/stackitcloud/stackit-sdk-go/core/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/sqlserverflex"
+	"github.com/stackitcloud/stackit-sdk-go/services/sqlserverflex/wait"
+)
+
+func main() {
+	// Specify the project ID
+	projectId := "16f49d71-37ad-4137-8b97-44d9c55c4094"
```
DiogoFerrao, line 16: This should be a placeholder

examples/sqlserverflex/sqlserverflex.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Specify instance configuration options
+	// flavorId := "FLAVOR_ID"
+	// version := "VERSION"
```
DiogoFerrao, line 20: I think we can uncomment these and delete the actual flavor and version below. This way we dont induce errors if the flavors/versions become deprecated.

examples/sqlserverflex/sqlserverflex.go
joaopalet: Ups, these were leftovers from testing. Didn't mean to commit them

examples/sqlserverflex/sqlserverflex.go
joaopalet: Same as above

DiogoFerrao (approved):

PR #564, opened 2024-06-12
Generator: Update SDK /services/resourcemanager

stackit-pipeline:
Generated from GitHub run [9483558850](https://github.com/stackitcloud/stackit-api-specifications/actions/runs/9483558850)


services/resourcemanager/configuration.go
```diff
@@ -22,12 +22,12 @@ func NewConfiguration() *config.Configuration {
 		Debug:         false,
 		Servers: config.ServerConfigurations{
 			{
-				URL:         "https://resource-manager.api.{region}stackit.cloud",
+				URL:         "https://resource-manager.api.stackit.cloud",
```
GokceGK, line 25: is {region} not needed?

DiogoFerrao: For global APIs the variable is not needed, and the generator just removes the variable and sets the default value int the ServerVariable config. For regional APIs it retains the variable (
[here](https://github.com/stackitcloud/stackit-sdk-go/blob/main/services/ske/configuration.go)
for instance)

GokceGK (approved):

PR #625, opened 2024-07-10
Add changelogs for fix of deprecation marks

joaopalet:


services/loadbalancer/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## v0.XX.X (2024-XX-XX)
+## v0.13.1 (2024-07-10)
```
DiogoFerrao, line 1: I'm not sure about this being a patch version, if the user has a linter like we do the builds will start failing.

I think this should be minor, since in the "Go way" we are only now effectively deprecating the methods

joaopalet: Good point

DiogoFerrao (approved):

PR #630, opened 2024-07-12
Add waiters for service enablement EnableService and DisableService

DiogoFerrao:


services/serviceenablement/wait/wait.go
```diff
@@ -0,0 +1,74 @@
...
+			return true, s, nil
+		case ServiceStateEnabling:
+			return false, nil, nil
+		case ServiceStateDisabled:
+			return true, s, fmt.Errorf("enabling failed for service with id %s", serviceId)
+		case ServiceStateDisabling:
+			return true, s, fmt.Errorf("service with id %s is in state %s", serviceId, *s.State)
+		}
+	})
+	handler.SetTimeout(45 * time.Minute)
```
joaopalet, line 46: Did you test it for some service? Don't we need a sleep before wait of some seconds or does the state immediatelly change to ENABLING after the enable call?

joaopalet (approved):

services/serviceenablement/wait/wait.go
DiogoFerrao: Good point, I only tested enabling an already enabled service. Disabling it failed.
Fixed now

joaopalet (approved):

PR #631, opened 2024-07-12
Add example usage for service enablement

DiogoFerrao:


examples/serviceenablement/serviceenablement.go
```diff
@@ -0,0 +1,68 @@
...
+	// List Services
+	listServicesResp, err := client.ListServiceStatusExecute(context.Background(), projectId)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error when calling `ListServices`: %v\n", err)
+	} else {
+		fmt.Printf("Number of services: %v\n", len(*listServicesResp.Items))
+	}
+
+	// Get Service Status
+	serviceId := "SERVICE_ID"
```
joaopalet, line 35: In the example we could get the `serviceId` from the response of the previous call, right?

DiogoFerrao: Yes we can, doing that

GokceGK (approved):

PR #634, opened 2024-07-18
Generator: Update SDK /services/runcommand

stackit-pipeline:
Generated from GitHub run [9990686317](https://github.com/stackitcloud/stackit-api-specifications/actions/runs/9990686317)


go.work
```diff
@@ -45,6 +45,7 @@ use (
 	./services/rabbitmq
 	./services/redis
 	./services/resourcemanager
+	./services/runcommand
```
DiogoFerrao, line 48: I don't think we need to add it to the go.work just yet. This could be done when we implement examples

GokceGK: If I remove it from go.work, linter is complaining that module not listed in go.work. As it has no disadvantage, i would keep it for now in go.work.

DiogoFerrao (approved):

PR #668, opened 2024-08-05
Implement waiter function for the create network by using the network id

GokceGK:


services/iaas/wait/wait.go
```diff
@@ -110,6 +110,27 @@ func CreateNetworkWaitHandler(ctx context.Context, a APIClientInterface, project
 	return handler
 }
 
+// CreateNetworkWaitHandlerWithNetworkId will wait for network creation using network id
+func CreateNetworkWaitHandlerWithNetworkId(ctx context.Context, a APIClientInterface, projectId, networkId string) *wait.AsyncActionHandler[iaas.Network] {
```
joaopalet, line 114: I think we could already replace the old waiter as it is not needed anymore, and then we can replace it in the TFP

joaopalet: Renaming this back to the correct name afterwards would be a breaking change anyway

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Release (2024-XX-XX)
 
+- `iaas`: [v0.6.0](services/iaas/CHANGELOG.md#v060-2024-08-05)
+  - **Breaking change:** Use network ID in the waiter: `CreateNetworkWaitHandler`
```
joaopalet, line 4:
```suggestion
  - **Breaking change:** Use network ID instead of request ID in the waiter: `CreateNetworkWaitHandler`
```

services/iaas/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v0.6.0 (2024-08-05)
+- **Breaking change:** Use network ID in the waiter: `CreateNetworkWaitHandler`
```
joaopalet, line 2:
```suggestion
- **Breaking change:** Use network ID instead of request ID in the waiter: `CreateNetworkWaitHandler`
```

joaopalet (approved):

joaopalet (approved):

PR #853, opened 2024-09-26
Add waiters for iaas volumes

vicentepinto98:


services/iaasalpha/wait/wait.go
```diff
@@ -0,0 +1,69 @@
...
+	"time"
+
+	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
+	"github.com/stackitcloud/stackit-sdk-go/core/wait"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaasalpha"
+)
+
+const (
+	AvailableStatus = "AVAILABLE"
+	DeleteSuccess   = "DELETED"
```
joaopalet, line 16: There should be other states right? Are they just not relevant for the waiting?

vicentepinto98: Not relevant. Updates are sync

joaopalet: I meant e.g. a `CREATING` state

vicentepinto98: There is a `CREATING` state. Even though we are doing that in other waiters, I don't think it adds much value to check for intermediate states, unless we want to cover all possible states but I think that's introducing complexity and a dependency, because the team might introduce other states at some point.

We could check for `ERROR` state to fail early though, I'll add that case

joaopalet (approved):

PR #865, opened 2024-10-02
IaaS Server waiters

vicentepinto98:


services/iaasalpha/wait/wait.go
```diff
@@ -15,11 +15,14 @@ const (
 	AvailableStatus = "AVAILABLE"
 	DeleteSuccess   = "DELETED"
 	ErrorStatus     = "ERROR"
+	ActiveStatus    = "ACTIVE"
+	ResizingStatus  = "RESIZING"
```
joaopalet, line 19: Are the two new ones only regarding Servers and the previous ones only regarding Volumes? If yes I think we could make it more clear by adding prefixes like `VolumeAvailableStatus`, `ServerActiveStatus`, etc

services/iaasalpha/wait/wait_test.go
```diff
@@ -36,6 +38,33 @@ func (a *apiClientMocked) GetVolumeExecute(_ context.Context, _, _ string) (*iaa
...
 }
 
+func (a *apiClientMocked) GetServerExecute(_ context.Context, _, _ string) (*iaasalpha.Server, error) {
+	if a.returnResizing {
+		a.returnResizing = false
+		return &iaasalpha.Server{
+			Id:     utils.Ptr("sid"),
+			Status: utils.Ptr(ResizingStatus),
+		}, nil
+	}
```
joaopalet, line 48: :+1:

services/iaasalpha/wait/wait.go
vicentepinto98: Good point

joaopalet (approved):

PR #873, opened 2024-10-10
Refactor Server ResizeWaitHandler

vicentepinto98:


services/iaasalpha/wait/wait.go
```diff
@@ -100,47 +100,33 @@ func CreateServerWaitHandler(ctx context.Context, a APIClientInterface, projectI
...
+				h.IntermediateStateReached = true
+				return false, server, nil
 			}
-			return true, server, fmt.Errorf("resizing failed for server with id %s", serverId)
+			if *server.Id == serverId && *server.Status == ErrorStatus {
+				if server.ErrorMessage != nil {
+					return true, server, fmt.Errorf("resizing failed for server with id %s: %s", serverId, *server.ErrorMessage)
+				}
+				return true, server, fmt.Errorf("resizing failed for server with id %s", serverId)
+			}
```
joaopalet, line 126: This check is already done below, I think we don't need it here

joaopalet: Could be moved to first place if it makes it easier to read

core/wait/wait.go
```diff
@@ -20,11 +20,12 @@ type AsyncActionCheck[T any] func() (waitFinished bool, response *T, err error)
...
-	sleepBeforeWait   time.Duration
-	throttle          time.Duration
-	timeout           time.Duration
-	tempErrRetryLimit int
+	checkFn                  AsyncActionCheck[T]
+	sleepBeforeWait          time.Duration
+	throttle                 time.Duration
+	timeout                  time.Duration
+	tempErrRetryLimit        int
+	IntermediateStateReached bool
```
joaopalet, line 28: There is no use case to reset the intermediate state to `false`, right?

joaopalet (approved):

PR #973, opened 2024-11-08
Generator: Update SDK /services/iaas

stackit-pipeline:
Generated from GitHub run [11744492443](https://github.com/stackitcloud/stackit-api-specifications/actions/runs/11744492443)


services/iaas/api_default.go
```diff
@@ -26,30 +26,30 @@ import (
 // DefaultApiService DefaultApi service
 type DefaultApiService service
 
-type ApiAddNICToServerRequest struct {
+type ApiAddNetworkToServerRequest struct {
```
DiogoFerrao, line 29: This one was also added right? I assume it's different from the `AddNicToServer` and adds a network with an id instead of a nic

joaopalet: Not really, if you notice it's already present in the previous version. Just that since the order changed GitHub does not do the best job identifying the actual changes

DiogoFerrao (approved):

PR #996, opened 2024-11-14
Generator: Update SDK /services/dns

stackit-pipeline:
Generated from GitHub run [11837544689](https://github.com/stackitcloud/stackit-sdk-generator/actions/runs/11837544689)


services/dns/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## v0.12.0 (2024-11-20)
+
+- **Feature:** New models `ZoneModelsImportRecordModel` and `ZoneModelsImportZoneJson`
```
DiogoFerrao, line 3: This `ZoneModelsImportRecordModel` is a rename of the `RecordDataExchange`, right?

GokceGK: In theory yes, they also have the same fields. But they did not remove the `RecordDataExchange`. So both of the models is still available in the service. That is why I did not call it renamed

PR #1034, opened 2024-11-27
Allow lint and test specific service

joaopalet:
- Allow `lint` and `test` a specific service via:
  - `$ make lint service=SERVICE_NAME`
  - `$ make test service=SERVICE_NAME`
- Default `make lint` and `make test` keeps previous behaviour of running for all services and `core`

### DIFF
--- a/scripts/test-go.sh
+++ b/scripts/test-go.sh
@@ -25,23 +25,25 @@ else
     exit 1
 fi
 
+test_service() {
+    service=$1
+
+    echo ">> Testing services/${service}"
+    if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
+        go test ${SERVICES_PATH}/${service}/... ${GOTEST_ARGS} # All manually maintained files are in subfolders
+    else
+        go test ${SERVICES_PATH}/${service}/... ${GOTEST_ARGS}
+    fi
+}
+
 # If a service is specified, only test that service
 if [ ! -z "${SERVICE}" ]; then
-    echo ">> Testing services/${SERVICE}"
-    cd ${SERVICES_PATH}/${SERVICE}
-    if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
-        go test ./ ${GOTEST_ARGS} # All manually maintained files are in subfolders
-    else
-        go test ./... ${GOTEST_ARGS}
-    fi
-    exit 0
-# Otherwise, test all modules
+    test_service ${SERVICE}
+# Otherwise, test all services and core
 else
-
     if [ "${SKIP_NON_GENERATED_FILES}" = false ]; then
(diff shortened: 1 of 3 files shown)


scripts/test-go.sh
```diff
@@ -21,27 +25,41 @@ else
...
+    echo ">> Testing services/${SERVICE}"
+    cd ${SERVICES_PATH}/${SERVICE}
     if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
         go test ./ ${GOTEST_ARGS} # All manually maintained files are in subfolders
     else
         go test ./... ${GOTEST_ARGS}
     fi
-done
+    exit 0
+# Otherwise, test all modules
```
DiogoFerrao, line 38:
```suggestion
# Otherwise, test all services
```

scripts/test-go.sh
```diff
@@ -21,27 +25,41 @@ else
...
-        echo ">> Skipping services/${service}"
-        continue
-    fi
-
-    echo ">> Testing services/${service}"
-    cd ${service_dir}
+# If a service is specified, only test that service
+if [ ! -z "${SERVICE}" ]; then
+    echo ">> Testing services/${SERVICE}"
+    cd ${SERVICES_PATH}/${SERVICE}
```
DiogoFerrao, line 31: I guess this was already a thing before, but we don't actually need to `cd` into the folders, we can just run `go test ${SERVICES_PATH}/${SERVICE}/... ${GOTEST_ARGS}`

scripts/test-go.sh
```diff
@@ -21,27 +25,41 @@ else
...
     if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
         go test ./ ${GOTEST_ARGS} # All manually maintained files are in subfolders
     else
         go test ./... ${GOTEST_ARGS}
     fi
-done
+    exit 0
```
DiogoFerrao, line 37: Why exit here and not in the `else`? Should be the same, right?

scripts/test-go.sh
```diff
@@ -21,27 +25,41 @@ else
...
-    cd ${service_dir}
+# If a service is specified, only test that service
+if [ ! -z "${SERVICE}" ]; then
+    echo ">> Testing services/${SERVICE}"
+    cd ${SERVICES_PATH}/${SERVICE}
     if [ "${SKIP_NON_GENERATED_FILES}" = true ]; then
         go test ./ ${GOTEST_ARGS} # All manually maintained files are in subfolders
     else
         go test ./... ${GOTEST_ARGS}
     fi
```
DiogoFerrao, line 36: Since this snippet is now duplicated, it may make sense to extract it into a `func`

DiogoFerrao (approved):

PR #1096, opened 2024-12-05
Add waiters and example for the image operations on `iaasalpha`

joaopalet:
- Adds waiters for two operations
  - `ImageUploadWaitHandler` waits for the image to become `AVAILABLE` after a successful upload of the image to the upload URL
  - DeleteImageWaitHandler` waits for the deletion of the image resource
- The actual creation of the image resource (done via `CreateImage`) is actually synchronous

### DIFF
--- a/examples/iaasalpha/image/image.go
+++ b/examples/iaasalpha/image/image.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/stackitcloud/stackit-sdk-go/core/config"
+	"github.com/stackitcloud/stackit-sdk-go/core/utils"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaasalpha"
+	"github.com/stackitcloud/stackit-sdk-go/services/iaasalpha/wait"
+)
+
+func main() {
+	// Specify the and project ID
+	projectId := "PROJECT_ID"
+	imageFilePath := "PATH/TO/IMAGE" // Should be a path to a valid image file, e.g. "./my-image.qcow2"
+	imageDiskFormat := "DISK_FORMAT" // E.g. "qcow2", "raw", "iso"
+
+	// Create a new API client, that uses default authentication and configuration
+	iaasalphaClient, err := iaasalpha.NewAPIClient(
+		config.WithRegion("eu01"),
+	)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Creating API client: %v\n", err)
+		os.Exit(1)
+	}
+	ctx := context.Background()
+
+	// Create an image
+	createImagePayload := iaasalpha.CreateImagePayload{
+		Name:       utils.Ptr("my-image"),
+		DiskFormat: utils.Ptr(imageDiskFormat),
+	}
+	imageCreateResp, err := iaasalphaClient.CreateImage(ctx, projectId).CreateImagePayload(createImagePayload).Execute()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when calling `CreateImage`: %v\n", err)
+		os.Exit(1)
+	} else {
+		fmt.Printf("[iaasalpha API] Image %q has been successfully created.\n", *imageCreateResp.Id)
+	}
+
+	// Upload the image by making a PUT request to upload URL
+	fileContents, err := os.ReadFile(imageFilePath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when reading file: %v\n", err)
+		os.Exit(1)
+	}
+
+	req, err := http.NewRequest(http.MethodPut, *imageCreateResp.UploadUrl, bytes.NewReader(fileContents))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when creating request: %v\n", err)
+		os.Exit(1)
+	}
+	req.Header.Set("Content-Type", "application/octet-stream")
+
+	fmt.Printf("[iaasalpha API] Uploading image %q...\n", *imageCreateResp.Id)
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
 	handler.SetTimeout(15 * time.Minute)
 	return handler
 }
+
+// ImageUploadWaitHandler will wait for server creation
```
vicentepinto98, line 353: We should add more info here on the process and what the function is actually checking: status AVAILABLE that indicates the upload of the image was successful

services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
...
+func ImageUploadWaitHandler(ctx context.Context, a APIClientInterface, projectId, imageId string) *wait.AsyncActionHandler[iaasalpha.Image] {
+	handler := wait.New(func() (waitFinished bool, response *iaasalpha.Image, err error) {
+		image, err := a.GetImageExecute(ctx, projectId, imageId)
+		if err != nil {
+			return false, image, err
+		}
+		if image.Id == nil || image.Status == nil {
+			return false, image, fmt.Errorf("create failed for image with id %s, the response is not valid: the id or the status are missing", imageId)
```
vicentepinto98, line 361:
```suggestion
			return false, image, fmt.Errorf("upload failed for image with id %s, the response is not valid: the id or the status are missing", imageId)
```

services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
...
+		}
+		if *image.Id == imageId && *image.Status == ImageAvailableStatus {
+			return true, image, nil
+		}
+		if *image.Id == imageId && *image.Status == ErrorStatus {
+			return true, image, fmt.Errorf("create failed for image with id %s", imageId)
```
vicentepinto98, line 367:
```suggestion
			return true, image, fmt.Errorf("upload failed for image with id %s", imageId)
```

services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
...
+		}
+		return false, image, nil
+	})
+	handler.SetTimeout(15 * time.Minute)
```
vicentepinto98, line 371: the timeout is highly dependent on the size of the image, but I would say in general we can increase this to 30 minutes

services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
...
+	return handler
+}
+
+// DeleteImageWaitHandler will wait for volume deletion
```
vicentepinto98, line 375:
```suggestion
// DeleteImageWaitHandler will wait for image deletion
```

services/iaasalpha/wait/wait.go
```diff
@@ -346,3 +349,53 @@ func DeleteVirtualIPWaitHandler(ctx context.Context, a APIClientInterface, proje
 	handler.SetTimeout(15 * time.Minute)
 	return handler
 }
+
+// ImageUploadWaitHandler will wait for server creation
+func ImageUploadWaitHandler(ctx context.Context, a APIClientInterface, projectId, imageId string) *wait.AsyncActionHandler[iaasalpha.Image] {
```
vicentepinto98, line 354: Why not `UploadImageWaitHandler`?
More uniform with the other waiter names

examples/iaasalpha/image/image.go
```diff
@@ -0,0 +1,99 @@
...
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when uploading image: %v\n", resp.Status)
+		os.Exit(1)
+	} else {
+		fmt.Printf("[iaasalpha API] Image %q has been successfully uploaded.\n", *imageCreateResp.Id)
+	}
+
+	// Wait for image to become available
+	image, err := wait.ImageUploadWaitHandler(ctx, iaasalphaClient, projectId, *imageCreateResp.Id).WaitWithContext(ctx)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when waiting for creation: %v\n", err)
```
vicentepinto98, line 77:
```suggestion
		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when waiting for upload: %v\n", err)
```

examples/iaasalpha/image/image.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	fmt.Printf("[iaasalpha API] Image %q is available.\n", *image.Id)
+
+	// Delete the image
+	err = iaasalphaClient.DeleteImage(ctx, projectId, *imageCreateResp.Id).Execute()
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "[iaasalpha API] Error when calling `DeleteImage`: %v\n", err)
+	} else {
+		fmt.Printf("[iaasalpha API] public IP %q has been successfully deleted.\n", *imageCreateResp.Id)
```
vicentepinto98, line 88:
```suggestion
		fmt.Printf("[iaasalpha API] Image %q has been successfully deleted.\n", *imageCreateResp.Id)
```

services/iaasalpha/wait/wait.go
joaopalet: Also though about it, did it differently since there is no actual operation called `UploadImage`. But we can keep the uniformity yes

services/iaasalpha/wait/wait.go
joaopalet: The PUT request to upload the image already blocks until the upload is complete, so I think in the normal case the waiter just waits for the state to change after the upload

vicentepinto98 (approved):

PR #1256, opened 2025-01-02
Region adjustment

Fyusel:
STACKIT will move to a new way of specifying regions with v1. Instead of providing the region in the URL directly it is provided via path parameter. Once all services have migrated, the methods to specify the region in the client configuration will be removed.

### DIFF
--- a/core/config/config.go
+++ b/core/config/config.go
@@ -5,6 +5,7 @@ package config
 import (
 	"context"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"strings"
@@ -478,7 +479,10 @@ func (c *Configuration) ServerURLWithContext(ctx context.Context, endpoint strin
 // ConfigureRegion configures the API server urls with the user specified region.
 // Does nothing if a custom endpoint is provided.
 // Throws an error if no region is given or if the region is not valid
+// Throws an error if a region is given for a global url.
 func ConfigureRegion(cfg *Configuration) error {
+	log.Println("WARNING: STACKIT will move to a new way of specifying regions, where the region is provided as a function argument instead of being set in the client configuration." +
+		" Once all services have migrated, the methods to specify the region in the client configuration will be removed.")
 	if cfg.setCustomEndpoint {
 		return nil
 	}
@@ -519,7 +523,12 @@ func ConfigureRegion(cfg *Configuration) error {
 		// Region is not available.
 		return fmt.Errorf("the provided region is not available for this API, available regions are: %s", availableRegions)
 	}
-	// Global API. The provided region is ignored.
+	// Global API.
+	// If a region is provided by the user via WithRegion() or via environment variable return an error.
+	// The region is provided as a function argument instead of being set in the client configuration.
+	if cfg.Region != "" {
+		return fmt.Errorf("this API does not support setting a region in the the client configuration, please check if the region can be specified as a function parameter")
+	}
 	// If the url is a template, generated using deprecated config.json, the region variable is replaced
 	// If the url is already configured, there is no region variable and it remains the same
 	cfgUrl := strings.Replace(servers[0].URL, "{region}", "", -1)
--- a/core/config/config_test.go
+++ b/core/config/config_test.go
@@ -113,7 +113,27 @@ func TestConfigureRegion(t *testing.T) {
 			isValid: true,
 		},
 		{
-			desc: "valid_deprecated_global_with_specified_region",
+			desc: "invalid_global_default_with_region",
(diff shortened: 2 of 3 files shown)


core/config/config.go
```diff
@@ -478,7 +478,10 @@ func (c *Configuration) ServerURLWithContext(ctx context.Context, endpoint strin
 // ConfigureRegion configures the API server urls with the user specified region.
 // Does nothing if a custom endpoint is provided.
 // Throws an error if no region is given or if the region is not valid
+// Throws an error if a region is given for a global url.
 func ConfigureRegion(cfg *Configuration) error {
+	fmt.Println("Warning: STACKIT will move to a new way of specifying regions with v1. Instead of providing the region in the URL directly it is provided via path parameter." +
```
joaopalet, line 483: We can use `log.Println` (from the standard library `"log"` package) instead of `fmt`

joaopalet:
```suggestion
	fmt.Println("WARNING: STACKIT will move to a new way of specifying regions, where the region is provided as a function argument instead of being set in the client configuration." +
```

core/config/config.go
```diff
@@ -519,13 +522,16 @@ func ConfigureRegion(cfg *Configuration) error {
...
-	// Global API. The provided region is ignored.
-	// If the url is a template, generated using deprecated config.json, the region variable is replaced
-	// If the url is already configured, there is no region variable and it remains the same
-	cfgUrl := strings.Replace(servers[0].URL, "{region}", "", -1)
+
+	// Global API. If a region is provided return an error.
+	// The region is not part of the url anymore instead the region is provided via path parameter.
+	if strings.Contains(servers[0].URL, "{region}") {
+		return fmt.Errorf("a region was provided in the url for a global api but has to be specified in the individual methods; STACKIT will move to a new way of specifying regions with v1")
+	}
```
joaopalet, line 530: We only want to return an error if the server URL does not contain a `{region}` placeholder

joaopalet: And the message can be something simple like "this API does not support setting a region in the server URL, please check if the region can be specified as a function parameter"

core/config/config.go
joaopalet: The suggestion also applies to the other places where this sentence is present

core/config/config.go
Fyusel: Are these two comments not contradictory? 
The error message of your second comment fits for if the URL contains a `{region}`.

core/config/config_test.go
```diff
@@ -113,7 +113,7 @@ func TestConfigureRegion(t *testing.T) {
 			isValid: true,
 		},
 		{
-			desc: "valid_deprecated_global_with_specified_region",
+			desc: "invalid_deprecated_global_with_specified_region",
```
joaopalet, line 116: Could you add another test case for a global server that is defined as
```
Servers: ServerConfigurations{
    ServerConfiguration{
        URL: "https://some-api.api.stackit.cloud",
            Variables: map[string]ServerVariable{
                "region": {
                    DefaultValue: "global",
                    EnumValues:   []string{},
                },
            },
    },
},
```

Fyusel: Added a test with (this combination was missing)
```
Region: "eu01",
Servers: ServerConfigurations{
	ServerConfiguration{
		URL: "https://some-api.api.stackit.cloud",
		Variables: map[string]ServerVariable{
			"region": {
				DefaultValue: "global",
				EnumValues:   []string{},
			},
		},
	},
},
```

core/config/config.go
```diff
@@ -478,7 +479,10 @@ func (c *Configuration) ServerURLWithContext(ctx context.Context, endpoint strin
 // ConfigureRegion configures the API server urls with the user specified region.
 // Does nothing if a custom endpoint is provided.
 // Throws an error if no region is given or if the region is not valid
+// Throws an error if a region is given for a global url.
 func ConfigureRegion(cfg *Configuration) error {
+	log.Println("WARNING: STACKIT will move to a new way of specifying regions, where the region is provided as a function argument instead of being set in the client configuration." +
```
joaopalet, line 484:
```suggestion
	slog.Warn("STACKIT will move to a new way of specifying regions, where the region is provided as a function argument instead of being set in the client configuration." +
```

joaopalet: Nitpick: this will already log with the log level `Warn`

Fyusel: log/slog is only available for Go 1.21+
Reverted the change.

joaopalet (approved):

PR #1262, opened 2025-01-08
Fix: ConfigureRegion() throws error if the region is set by environment variable

marceljk:


vicentepinto98 (commented):
this would require a new bugfix version of the core, so you should already update the CHANGELOG files.

The change looks ok :+1:

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Release (2025-01-08)
```
vicentepinto98, line 1: You don't need to always make a general GitHub release for each small change, so usually we set this date to something like "2025-XX-XX" and do the release around every 2 weeks

vicentepinto98 (approved):

vicentepinto98 (approved):

PR #1306, opened 2025-01-22
Fix test runner for sdk

bahkauv70:
Encapsulate the retrieval of the home directory. Otherwise the testcases will start to interfere with the real users home directory (this cannot be redefined platform-agnostic within go). The cause of this issue is, that service account keys are automatically searched in $HOME/.stackit/credentials.json which interferes with the testcases

### DIFF
--- a/core/auth/auth_test.go
+++ b/core/auth/auth_test.go
@@ -456,6 +456,8 @@ func TestKeyAuth(t *testing.T) {
 			configuredPrivateKey: string(configuredPrivateKey),
 			expectedPrivateKey:   string(configuredPrivateKey),
 			isValid:              true,
+			homeDir:              func(t *testing.T) string { return t.TempDir() },
+
 		},
 		{
 			desc:               "included_private_key",
@@ -463,6 +465,8 @@ func TestKeyAuth(t *testing.T) {
 			includedPrivateKey: &includedPrivateKey,
 			expectedPrivateKey: includedPrivateKey,
 			isValid:            true,
+			homeDir:              func(t *testing.T) string { return t.TempDir() },
+
 		},
 		{
 			desc:                 "empty_configured_use_included_private_key",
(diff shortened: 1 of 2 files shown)


core/auth/auth_test.go
```diff
@@ -484,6 +485,7 @@ func TestKeyAuth(t *testing.T) {
 			serviceAccountKey:    nil,
```
Fyusel, line 485: Cannot comment to above functions therefore I add the comments here:
Line 466 (test "included_private_key") also need to add:
`homeDir:            func(t *testing.T) string { return t.TempDir() },`

Same for test "empty_configured_use_included_private_key"

core/auth/auth_test.go
```diff
@@ -484,6 +485,7 @@ func TestKeyAuth(t *testing.T) {
...
 			configuredPrivateKey: configuredPrivateKey,
 			isValid:              false,
+			homeDir:              func(t *testing.T) string { return t.TempDir() },
 		},
 		{
 			desc:                 "missing_private_key",
```
Fyusel, line 491: For this test case it is also needed to add:
`homeDir:              func(t *testing.T) string { return t.TempDir() },` 
otherwise my `credentials.json` file is read

core/auth/auth_test.go
```diff
@@ -769,6 +781,7 @@ func TestGetServiceAccountKey(t *testing.T) {
 		credentialsFilePath         string
```
Fyusel, line 781: Same here cannot comment on the actual line: 
Test ( "no_private_key_provided") also needs the home dir adjustment you made in order to not read from the credentials.json (even if "" is provided my credentials.json is read)

core/auth/auth_test.go
bahkauv70: Fixed

core/auth/auth_test.go
bahkauv70: Fixed

Fyusel: This is still the case on my side.

Fyusel (approved):

PR #1499, opened 2025-02-25
Fix/key auth environment variables

### LINKED ISSUE #1303: Missing handling of some authentication environment variables
Hey Dev Tools team, in the process of both integrating the STACKIT Go SDK into our tools and then suggesting an improvement to the Authentication documentation (see #1302), I found that some of the environment variables that are supposed to be used for authentication are not actually being processed.

In https://github.com/stackitcloud/stackit-sdk-go/blob/2b7f634f11d0908c3033663658342f2865cf0512/core/auth/auth.go#L67C1-L75C80 the documentation of the `DefaultAuth` function mentions the usage of the `STACKIT_SERVICE_ACCOUNT_KEY` and `STACKIT_PRIVATE_KEY` environment variables, but these are not actually checked or used in the code.

You can see that in the [bit of code](https://github.com/stackitcloud/stackit-sdk-go/blob/2b7f634f11d0908c3033663658342f2865cf0512/core/auth/auth.go#L298C1-L306C2) that fetches both the service account key and the private key, only the environment variables that use the path are checked. I believe that these methods need to be extended to also support proving both the service account key and the private key directly, which is very useful in automated pipelines for instance, where providing a filepath is not practical.

Thank you!

Fyusel:
fixes #1303

### DIFF
--- a/core/auth/auth.go
+++ b/core/auth/auth.go
@@ -18,12 +18,16 @@ type Credentials struct {
 	STACKIT_SERVICE_ACCOUNT_TOKEN    string
 	STACKIT_SERVICE_ACCOUNT_KEY_PATH string
 	STACKIT_PRIVATE_KEY_PATH         string
+	STACKIT_SERVICE_ACCOUNT_KEY      string
+	STACKIT_PRIVATE_KEY              string
 }
 
 const (
 	credentialsFilePath                                = ".stackit/credentials.json" //nolint:gosec // linter false positive
 	tokenCredentialType                 credentialType = "token"
+	serviceAccountKeyCredentialType     credentialType = "service_account_key"
 	serviceAccountKeyPathCredentialType credentialType = "service_account_key_path"
+	privateKeyCredentialType            credentialType = "private_key"
 	privateKeyPathCredentialType        credentialType = "private_key_path"
 )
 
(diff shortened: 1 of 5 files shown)


core/auth/auth.go
```diff
@@ -268,21 +282,36 @@ func getServiceAccountEmail(cfg *config.Configuration) string {
 }
 
 // getKey searches for a key in the following order: client configuration, environment variable, credentials file.
-func getKey(cfgKey, cfgKeyPath *string, envVar, credType credentialType, cfgCredFilePath string) error {
+func getKey(cfgKey, cfgKeyPath *string, envVarKeyPath, envVarKey, credTypePath, credTypeKey credentialType, cfgCredFilePath string) error {
```
marceljk, line 285:
```suggestion
func getKey(cfgKey, cfgKeyPath *string, envVarKeyPath, envVarKey string, credTypePath, credTypeKey credentialType, cfgCredFilePath string) error {
```
`envVarKeyPath` and `envVarKey` aren't really of type `credentialType`, so they both can set as string. With that you also don't need to parse them into string in the calls `os.LookupEnv`

marceljk (dismissed):

marceljk (approved):

PR #1516, opened 2025-03-03
Feat: increase timeouts for iaas handlers

h3adex:
In this PR, I've adjusted the default timeout handlers for creating IaaS resources to accommodate the STACKIT Terraform provider's requirements. I've increased the timeouts for longer creation times, such as when provisioning volumes for Windows servers, which typically exceed 10 minutes.


Fyusel (commented):
Thank you for contributing!
Can you add a changelog entry as well please? (one in the service and one in the main changelog file)

Fyusel (approved):

PR #1627, opened 2025-03-15
Accept custom transport of custom http client

jkroepke:
Hi,

I'm looking into setting up the discovery integration for STACKIT in [prometheus/prometheus](https://github.com/prometheus/prometheus).

While the integration generally works fine, I encountered some issues injecting the http.RoundTripper from the Prometheus ecosystem into the SDK CLI.

Ref: https://github.com/jkroepke/prometheus/blob/8bec5e7cce575b18af42fe8cd67ee0fd5f47ac73/discovery/stackit/server.go#L91

It seems that each authentication method creates a new http.Client instead of simply extending the chain of RoundTripper functions.

This PR modifies the behavior so that the auth function accepts an existing http.RoundTripper and extends it, as is usual in the Go ecosystem.

Theoretically, this avoids the need for proprietary middleware approaches, as existing middleware can be injected through the Transport of a custom http.Client. That's how it usually works in the Go ecosystem.


bahkauv70:
hello @jkroepke , thanks for your contribution! I think this would be a good idea, but the current PR breaks backwards compatibility. In the present form, all projects using the SDK would have to be adapted to the changed api. In order to go forward with this, the compatibility to existing projects would have to be maintained

jkroepke:
Hi @bahkauv70 

the current approach should preserve the backwards compatibility.

marceljk:
Hi @jkroepke,

Thanks for your contribution! The PR looks good. Only a nitpick, can you add some changelogs to these two files:

https://github.com/stackitcloud/stackit-sdk-go/blob/main/core/CHANGELOG.md
https://github.com/stackitcloud/stackit-sdk-go/blob/main/CHANGELOG.md

jkroepke:
Hi,

I added some test case to increase the test coverage. At least with that tests, I feel a bit more confident that everything still works.

marceljk:
Hi,

there are some linter issues which you can check [here](https://github.com/stackitcloud/stackit-sdk-go/actions/runs/13998990153/job/39200908945?pr=1627) in the pipeline or you can run `make lint` to see them local

marceljk (approved):
Thanks for your contribution 😃

PR #1815, opened 2025-03-31
Feat/stackitsdk 72 onboard cdn api

bahkauv70:


services/cdn/wait/wait_test.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+			ctx := context.Background()
+			handler := CreateDistributionPoolWaitHandler(ctx, &apiClientMock, tt.args.projectId, tt.args.distributionId)
+			handler.SetTimeout(2 * time.Second)
+			handler.SetSleepBeforeWait(100 * time.Millisecond)
+			handler.SetThrottle(50 * time.Millisecond)
+			response, err := handler.WaitWithContext(ctx)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("handler error %s, wantErr %v", err, tt.wantErr)
+			}
```
marceljk, line 138: I think a return would make here more sense
```suggestion
			if (err != nil) != tt.wantErr {
				t.Fatalf("handler error %s, wantErr %v", err, tt.wantErr)
				return
			}
```

marceljk (approved):

services/cdn/wait/wait_test.go
bahkauv70: I could add it, but the fatalf will abort the testcase anyways

PR #1823, opened 2025-04-04
add wait handlers for custom domains and distributions

malteehrlen:
## Description

Add wait handlers for CDN resources https://jira.schwarz/browse/STACKITCDN-716

I just saw that a wait handler was already added, please consider this contribution as extending the existing handler with handlers also for custom domains

## Checklist

- [x] Issue was linked above
- [x] Changelogs

### DIFF
--- a/services/cdn/wait/wait.go
+++ b/services/cdn/wait/wait.go
@@ -74,8 +74,14 @@ func UpdateDistributionWaitHandler(ctx context.Context, api APIClientInterface, 
 			switch *distribution.Distribution.Status {
 			case DistributionStatusActive:
 				return true, distribution, err
+			case DistributionStatusUpdating:
+				return false, nil, nil
+			case DistributionStatusDeleting:
+				return true, nil, fmt.Errorf("updating CDN distribution failed")
+			case DistributionStatusError:
+				return true, nil, fmt.Errorf("updating CDN distribution failed")
 			default:
-				return false, distribution, err
+				return true, nil, fmt.Errorf("UpdateDistributionWaitHandler: unexpected status %s", *distribution.Distribution.Status)
 			}
 		}
 
@@ -88,16 +94,19 @@ func UpdateDistributionWaitHandler(ctx context.Context, api APIClientInterface, 
 
 func DeleteDistributionWaitHandler(ctx context.Context, api APIClientInterface, projectId, distributionId string) *wait.AsyncActionHandler[cdn.GetDistributionResponse] {
 	handler := wait.New(func() (waitFinished bool, distribution *cdn.GetDistributionResponse, err error) {
-		distribution, err = api.GetDistributionExecute(ctx, projectId, distributionId)
+		_, err = api.GetDistributionExecute(ctx, projectId, distributionId)
 
 		// the distribution is still gettable, e.g. not deleted
 		if err == nil {
 			return false, nil, nil
 		}
-		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
-		if ok && oapiErr.StatusCode == http.StatusNotFound {
-			return true, nil, nil
+		var oapiError *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiError) {
+			if statusCode := oapiError.StatusCode; statusCode == http.StatusNotFound || statusCode == http.StatusGone {
+				return true, nil, nil
+			}
 		}
+
 		return false, nil, err
 	})
 	handler.SetTimeout(30 * time.Second)
@@ -139,9 +148,11 @@ func DeleteCDNCustomDomainWaitHandler(ctx context.Context, a APIClientInterface,
 		if err == nil {
 			return false, nil, nil
 		}
-		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
-		if ok && oapiErr.StatusCode == http.StatusNotFound {
-			return true, nil, nil
+		var oapiError *oapierror.GenericOpenAPIError
+		if errors.As(err, &oapiError) {
+			if statusCode := oapiError.StatusCode; statusCode == http.StatusNotFound || statusCode == http.StatusGone {
+				return true, nil, nil
+			}
 		}
 		return false, nil, err
 	})
--- a/services/cdn/wait/wait.go
+++ b/services/cdn/wait/wait.go
@@ -13,16 +13,17 @@ import (
 )
 
 const (
-	DistributionStateCreating = "CREATING"
-	DistributionStateActive   = "ACTIVE"
-	DistributionStateUpdating = "UPDATING"
-	DistributionStateDeleting = "DELETING"
-	DistributionStateError    = "ERROR"
+	DistributionStatusCreating = "CREATING"
+	DistributionStatusActive   = "ACTIVE"
+	DistributionStatusUpdating = "UPDATING"
+	DistributionStatusDeleting = "DELETING"
+	DistributionStatusError    = "ERROR"
 )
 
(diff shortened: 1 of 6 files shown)


services/cdn/wait/wait.go
```diff
@@ -13,16 +13,17 @@ import (
 )
 
 const (
-	DistributionStateCreating = "CREATING"
-	DistributionStateActive   = "ACTIVE"
-	DistributionStateUpdating = "UPDATING"
-	DistributionStateDeleting = "DELETING"
-	DistributionStateError    = "ERROR"
+	DistributionStatusCreating = "CREATING"
```
bahkauv70, line 16: We have to be careful with renaming exported constants, as this might break our own and client code. In this case it is ok, as the constants are probably not used right now, so can do this.

services/cdn/wait/wait.go
```diff
@@ -39,17 +40,21 @@ func CreateDistributionPoolWaitHandler(ctx context.Context, api APIClientInterfa
 		}
 		if *distribution.Distribution.Id == distributionId {
 			switch *distribution.Distribution.Status {
-			case DistributionStateActive:
-				return true, distribution, err
+			case DistributionStatusActive:
+				return true, distribution, nil
+			case DistributionStatusCreating, DistributionStatusUpdating:
+				return false, nil, nil
+			case DistributionStatusDeleting:
```
bahkauv70, line 47: Honest question: Is it possible that a distribution goes into state "deleting" immediately after creation?

services/cdn/wait/wait.go
```diff
@@ -39,17 +40,21 @@ func CreateDistributionPoolWaitHandler(ctx context.Context, api APIClientInterfa
...
+				return true, nil, fmt.Errorf("creating CDN distribution failed")
```
bahkauv70, line 48: The original error should be contained within the returned error to facilitate debugging:

```suggestion
				return true, nil, fmt.Errorf("creating CDN distribution failed: %w",err)
```
This affects the following statements as well

services/cdn/wait/wait.go
```diff
@@ -84,17 +89,62 @@ func UpdateDistributionWaitHandler(ctx context.Context, api APIClientInterface,
 func DeleteDistributionWaitHandler(ctx context.Context, api APIClientInterface, projectId, distributionId string) *wait.AsyncActionHandler[cdn.GetDistributionResponse] {
 	handler := wait.New(func() (waitFinished bool, distribution *cdn.GetDistributionResponse, err error) {
 		distribution, err = api.GetDistributionExecute(ctx, projectId, distributionId)
+
+		// the distribution is still gettable, e.g. not deleted
+		if err == nil {
+			return false, nil, nil
+		}
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
```
bahkauv70, line 97: We should use this pattern instead:
```
			var oapiError *oapierror.GenericOpenAPIError
			if errors.As(err, &oapiError) {
				if statusCode := oapiError.StatusCode; statusCode == http.StatusNotFound || statusCode == http.StatusGone {
					return true, distribution, nil
				}
			}
```
which handles nested errors correctly. Additionally we found, that it is better to check for state 410 Gone as well. Some endpoints return this and we decided to check for it in all waiters.

services/cdn/wait/wait.go
```diff
@@ -84,17 +89,62 @@ func UpdateDistributionWaitHandler(ctx context.Context, api APIClientInterface,
...
-				}
-			}
+			return false, nil, err
+		}
+		if resp == nil || resp.CustomDomain == nil || resp.CustomDomain.Status == nil {
+			return false, nil, errors.New("CDNDistributionWaitHandler: status or custom domain missing in response")
+		}
+
+		switch *resp.CustomDomain.Status {
+		case cdn.DOMAINSTATUS_ACTIVE:
```
bahkauv70, line 118: See above regarding nested error messages

services/cdn/wait/wait.go
malteehrlen: yes, if the delete endpoint is called before creation is finished, which is 100% possible as it takes some 20 seconds to create it

services/cdn/wait/wait.go
```diff
@@ -84,17 +89,62 @@ func UpdateDistributionWaitHandler(ctx context.Context, api APIClientInterface,
...
-	handler.SetTimeout(10 * time.Minute)
+func DeleteCDNCustomDomainWaitHandler(ctx context.Context, a APIClientInterface, projectId, distributionId, domain string) *wait.AsyncActionHandler[cdn.CustomDomain] {
+	handler := wait.New(func() (waitFinished bool, response *cdn.CustomDomain, err error) {
+		_, err = a.GetCustomDomainExecute(ctx, projectId, distributionId, domain)
+
+		// the custom domain is still gettable, e.g. not deleted
+		if err == nil {
+			return false, nil, nil
+		}
+		oapiErr, ok := err.(*oapierror.GenericOpenAPIError) //nolint:errorlint //complaining that error.As should be used to catch wrapped errors, but this error should not be wrapped
```
bahkauv70, line 142: See above regarding the type check of the open api error

services/cdn/wait/wait.go
malteehrlen: there was no error here, `err` is nil...

services/cdn/wait/wait_test.go
```diff
@@ -3,27 +3,327 @@ package wait
...
 	"testing"
 	"time"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
 	"github.com/stackitcloud/stackit-sdk-go/services/cdn"
 )
 
-var _ APIClientInterface = (*apiClientMock)(nil)
+type mockApiClient struct {
```
bahkauv70, line 15: We should avoid using two different mocks. I do not have to much preference, _which_ mock we continue to use, but it should be a single one

services/cdn/wait/wait_test.go
```diff
@@ -3,27 +3,327 @@ package wait
...
+			dist, err := handler.SetTimeout(10 * time.Millisecond).WaitWithContext(context.Background())
+			tc.errCheck(t, err)
+			if diff := cmp.Diff(tc.expectedDistribution, dist); diff != "" {
+				t.Fatalf("Unexpected response (-want, +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestDeleteDistributionWaitHandler(t *testing.T) {
```
bahkauv70, line 180: s. above

services/cdn/wait/wait.go
malteehrlen: `err` is nil, there is nothing to nest

services/cdn/wait/wait.go
malteehrlen: done

services/cdn/wait/wait_test.go
```diff
@@ -3,27 +3,327 @@ package wait
...
+				}, nil
+			}
+		}
+	}
+	return nil, &oapierror.GenericOpenAPIError{
+		StatusCode: http.StatusNotFound,
+	}
+}
+
+func TestCreateDistributionWaitHandler(t *testing.T) {
```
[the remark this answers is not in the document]
malteehrlen: I decided to keep the new tests. Its a good proof that both the waiters were implemented similarly. Also, they use the existing mock

services/cdn/wait/wait_test.go
malteehrlen: the old tests seem to test mostly the wait loop which is not the code that I want to test here, I specifically want to test the logic in wait.go, e.g. not the retries and stuff that is handled elsewhere. So I decided to keep the new tests, it has the benefit that style is consistent

bahkauv70 (approved):

bahkauv70 (approved):

Issue #731, opened 2024-08-12
error in keyflow's `RoundTrip` is misleading


hown3d:
We ran into an issue the other day to debug why requests to an API endpoint fail.
The SDK was configured to use the key flow authentication but failed to get an access token for the configured STACKIT service account key. The error looked something like this:
```
URL: "https://ske.api.eu01.dev.stackit.cloud/v1/projects/XXXXX-96d3-XXXX-a2a2-5eff15f63074/clusters/e2e-juqgdz4",
Err: <*fmt.wrapError | 0xc0004af380>{
     msg: "get new access token: , status code 400, Body: {\"timeStamp\":\"2024-08-08T14:37:03.450943101\",\"path\":\"/token\",\"status\":400,\"error\":\"Bad Request\",\"message\":\"Key with kid XXXXX-a8fb-4c18-ae5e-XXXXX was not found\"}\n",
     err: <*oapierror.GenericOpenAPIError | 0xc000884440>{
          StatusCode: 400,
          Body: "{\"timeStamp\":\"2024-08-08T14:37:03.450943101\",\"path\":\"/token\",\"status\":400,\"error\":\"Bad Request\",\"message\":\"Key with kid XXXXX-a8fb-4c18-ae5e-XXXXX was not found\"}",
          ErrorMessage: "",
          Model: nil,
      },
 },
```

The error returned is really misleading since the key flow authentication just returns a generic error in `RoundTrip` if the retrieval of an access token fails. 
https://github.com/stackitcloud/stackit-sdk-go/blob/0c4f21ec4c53140050ac6877389d3fdd508d9974/core/clients/key_flow.go#L165-L176

I would suspect to get a more meaningful error, maybe even a typed one, to look out for authentication problems.

GokceGK:
Hi @hown3d, 

thanks a lot for getting in touch with us and reporting the issue. We will verify the enhancement possibilities and apply them accordingly. We will inform you here about the progress.

marceljk:
Hi @hown3d,

we improved the error message for the keyflow RoundTrip and it's now available in the version v0.17.1 of core module.

PR #1846, opened 2025-04-09
feat(core): improve roundtripper error message

marceljk:
## Description

relates to #731 and STACKITSDK-53

## Checklist

- [x] Issue was linked above
- [x] Changelogs

### DIFF
--- a/core/clients/key_flow.go
+++ b/core/clients/key_flow.go
@@ -6,10 +6,12 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


core/clients/key_flow.go
```diff
@@ -209,7 +211,14 @@ func (c *KeyFlow) GetAccessToken() (string, error) {
 	if !accessTokenExpired {
 		return accessToken, nil
 	}
-	if err := c.recreateAccessToken(); err != nil {
+	if err = c.recreateAccessToken(); err != nil {
+		var oapiErr *oapierror.GenericOpenAPIError
+		if ok := errors.As(err, &oapiErr); ok {
+			reg := regexp.MustCompile("Key with kid ([a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}) was not found")
```
bahkauv70, line 217: As the matched UUID is not really used, the pattern could probably replaced with a simple .*? (non greedy match)

bahkauv70: Note for future extension: If we start checking multiple error messages here, we should probably think about a tabular definition of regexp -> user presentable error message. Perhaps we could even discuss with the team responsible for the auth stuff to provide us with unique error codes...

marceljk: I'll ask the team if they can provide us with all the error codes. Because in the API spec there is currently only the 200 response defined https://docs.api.eu01.stackit.cloud/documentation/service-account/version/v2#tag/Public/operation/post-token

bahkauv70 (approved):

PR #2179, opened 2025-05-08
core: Add STACKIT CLI Auth flow

jkroepke:
## Description

relates to https://github.com/stackitcloud/terraform-provider-stackit/issues/719

This PR adds the ability to obtain a token from a locally installed CLI. This is great for local development, where a real user is logged in via the CLI, and other STACKIT-related tools like the Terraform provider can use the token from the authenticated CLI context to make additional requests against the STACKIT API.

While developing this feature, I found https://github.com/stackitcloud/stackit-cli/pull/736.

Due limited STACKIT access, tests with real human accounts can not be done by myself.

The Stackit CLI flow is enabled by default. There is an option to disable the cli flow. On CLI itself, the option should set to false to prevent infinite loops.

The path of the STACKIT CLI is not configurable. The SDK expects the CLI to be available within the system PATH. I modified the Github Action to ensure that the Stackit CLI is present.

Tests are using a custom CLI profile to avoid interrupting the local default configuration.

Code inspired from https://github.com/Azure/azure-sdk-for-go/blob/f4b3a417e0bc39a5051028c4f9d0fc36bc612317/sdk/azidentity/azure_cli_credential.go#L119-L166

## Checklist

- [ ] Issue was linked above
- [x] Changelogs


.github/workflows/ci.yaml
```diff
@@ -17,6 +17,8 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version: ${{ matrix.go-version }}
+      - name: Install STACKIT CLI
+        uses: jkroepke/setup-stackit-cli@v1
```
BackInBash, line 21: We may want to fork this to keep dependencies clean.

But I like to get other opinions on that.

BackInBash (commented):
GitHub Workflow dependency check

.github/workflows/ci.yaml
jkroepke: I could use a simple curl here as well.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Release (2025-XX-YY)
+- `core`: [v0.17.2](core/CHANGELOG.md#v0172-2025-XX-YY)
+  - **New:** If a custom http.Client is provided, the http.Transport is respected. This allows customizing the http.Client with custom timeouts or instrumentation.
```
jkroepke, line 3:
```suggestion
  - **New:** Add STACKIT CLI auth flow.
```

.github/workflows/ci.yaml
jkroepke, line 21:
```suggestion
      - name: Install STACKIT CLI
        run: >-
          OS=$(uname | tr '[:upper:]' '[:lower:]')
          ARCH=$(uname -m)
          
          [ "$ARCH" = "x86_64" ] && ARCH=amd64
          [ "$ARCH" = "aarch64" ] && ARCH=amd64
          
          curl -sSfLo - https://github.com/stackitcloud/stackit-cli/releases/download/v0.31.0/stackit-cli_0.31.0_${OS}_${ARCH}.tar.gz | sudo tar zxvf - -C /usr/local/bin/ stackit
```

core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v0.17.2 (2025-XX-YY)
```
marceljk, line 1:
```suggestion
## v0.17.2 (2025-05-21)
```

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## Release (2025-XX-YY)
+- `core`: [v0.17.2](core/CHANGELOG.md#v0172-2025-XX-YY)
```
marceljk, line 2:
```suggestion
- `core`: [v0.17.2](core/CHANGELOG.md#v0172-2025-05-21)
```

marceljk (commented):
Looks very good and works as expected. 😃 Can you just update the date of the changelog?

jkroepke:
@marceljk Done and thanks for the additional testing.

Keep in mind to set `DisableCLIAuthFlow: true` in the STACKIT CLI to avoid auth loops.

jkroepke:
Hi @marceljk 

I would have a different sight on that:

On the Azure SDK, it's enabled by default. This at least guarantees that CLI authentication is always available, even for third-party programs like Prometheus, Terraform. And it's documented:
https://learn.microsoft.com/en-us/azure/developer/go/sdk/authentication/authentication-overview

The AWS SDK works in a similar way. An authenticated CLI is sufficient for the Terraform provider.

From an end-user perspective, I would assume that CLI authentication is enabled by default.

But I respect your decision and have changed it as requested.

marceljk:
Hi @jkroepke,

that's a good point that the CLI authentication is by default always available and even for third-party programs. We will discuss it again internal, what's the better option here and will come back to you.

marceljk:
Hello @jkroepke,

Thanks for your contribution! We really appreciated your contribution and this gave us some new input for our internal discussion and how we can implement this feature. 
We decided, that we want to make also some changes in the CLI and add new commands for this purpose. So users would have better control if the CLI Auth flow is used or not and we gain more control over the used interface in the CLI.

We created a RFC https://github.com/stackitcloud/terraform-provider-stackit/discussions/880 with our target architecture of this feature. We would appreciate it, when you give us some feedback to it. When we implement this feature, we can also add you as reviewer if you like to.
 
With this newly defined architecture, I will close this PR.

PR #2194, opened 2025-05-12
fix(key_flow): Add 5 second leeway to refresh access tokens early

### LINKED ISSUE #2193: Access Tokens generated in Service Account Key client can be used too close to expiration time
## Description

Access tokens generated via the service account key flow may expire during the time from validating the tokens expiration time until it reaches the upstream systems validating the token.

## Steps to reproduce

Due to the hard coded default token lifetime of 10 minutes (and no way to access it from SDK external code to wait an accurate amount of time) this is hard to reproduce in a code sample.

We have reconstructed this behavior as the most likely cause from user reports and internal logs.
Please get in touch with me if you need further insights.

## Actual behavior

If an access token is about to expire but still used for an upstream request, authentication systems may respond with an unexpected `401 Unauthorized` resonse:
`{\"status\":401,\"message\":\"Unauthenticated token\"}`

## Expected behavior

SDK should refresh access tokens early if they are almost expired to prevent such issues.

## Environment
 - OS: macOS 15.4.1
 - Go version (see `go version`): `1.24.2`
 - Version of the STACKIT Go SDK:  `v0.16.2`

```
$ cat go.mod | grep "stackit"

github.com/stackitcloud/stackit-sdk-go/core v0.16.2
github.com/stackitcloud/stackit-sdk-go/services/cdn v0.3.0
github.com/stackitcloud/stackit-sdk-go/services/resourcemanager v0.12.0
github.com/stackitcloud/stackit-sdk-go/services/dns v0.12.1
github.com/stackitcloud/stackit-sdk-go/services/postgresflex v0.17.0
```

**Additional information**

It is in general good practice to have some leeway when working with timestamps in access tokens, especially to compensate for the common scenario of clock skew between client/server systems.

ptrckmllr:
## Description

This PR attempts to fix #2193

## Checklist

- [x] Issue was linked above
- [x] Changelogs


ptrckmllr:
It seems this breaks the `key_flow_continuous_refresh_test.go` tests due to the extremly small token lifetimes used in there.

Is increasing token lifetimes (5+ seconds) in those tests a feasible approach?

malteehrlen:
thanks for this, we have been suffering expired tokens every now and then <3

ptrckmllr:
Moved the new variable to `KeyFlow` struct, set it during `Init(...)` and validate it.

Now the `key_flow_continuous_refresh_test.go` do work as expected again, as it uses the default 0 value.

This raises questions about the robustness of these tests though, as they do not call the `KeyFlow.Init(...)` func and [instead just create the `KeyFlow` struct](https://github.com/stackitcloud/stackit-sdk-go/blob/cf32f726358dc5c468822e9d8181ba13fd35c7c5/core/clients/key_flow_continuous_refresh_test.go#L136).

Should I create a separate issue or is this the intended way for the tests to run?

ptrckmllr:
> Thank you for contributing! Looks good, do you want to add the changelogs as well then this can be merged.
> 
> Regarding your open topic, I created a separate ticket on our side to have a look at this.

Changelog has been updated.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Release (2025-YY-XX)
+- core:
```
Fyusel, line 2:
```suggestion

```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+  - v0.17.2
```
Fyusel, line 3:
```suggestion
- `core`: [v0.17.2](core/CHANGELOG.md#v0172-2025-05-22)
```

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+    - Bugfix: Access tokens generated via key flow authentication are refreshed 5 seconds before expiration to prevent timing issues with upstream systems which could lead to unexpected 401 error responses
```
Fyusel, line 4:
```suggestion
  - **Bugfix:** Access tokens generated via key flow authentication are refreshed 5 seconds before expiration to prevent timing issues with upstream systems which could lead to unexpected 401 error responses
```

core/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## v0.17.2 (2025-05-22)
+- Bugfix: Access tokens generated via key flow authentication are refreshed 5 seconds before expiration to prevent timing issues with upstream systems which could lead to unexpected 401 error responses
```
Fyusel, line 2:
```suggestion
- **Bugfix:** Access tokens generated via key flow authentication are refreshed 5 seconds before expiration to prevent timing issues with upstream systems which could lead to unexpected 401 error responses
```

Fyusel (approved):

PR #3375, opened 2025-07-28
fix: update service tags fails, when .openapi-generator/VERSION was changed

marceljk:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above
- [ ] Changelogs


scripts/update-service-tags.sh
```diff
@@ -8,6 +8,12 @@ set -e
 
 # Check all version files which have changed
 for file in $(git diff --name-only HEAD~1..HEAD | grep VERSION); do
```
rubenhoenle:
```suggestion
for file in $(git diff --name-only HEAD~1..HEAD | grep "services/[^/]\+/VERSION$"); do
```

Why not adjust the grep command to do exactly what we want?

Some explanation: The `[^/]\+` part in the middle matches everything except slashes. So the nesting is maxed to 1 level.

Try it yourself: Write some test strings into a file (or do some git commit on a local branch) and run either
- `git diff --name-only HEAD~1..HEAD | grep "services/[^/]\+/VERSION$"`
- `cat input.txt | grep "services/[^/]\+/VERSION$"`

This should match the following ones (`+` is match, `-` means no match):
(+) services/cdn/VERSION
(-) services/cdn/sub/VERSION
(-) services/cdn/.openapi-generator/VERSION
(+) services/dns/VERSION

rubenhoenle: We could also use something like `grep "services/[A-Za-z0-9\-_]\+/VERSION$"` if you prefer in terms of readability

rubenhoenle: wait, it still has to work for https://github.com/stackitcloud/stackit-sdk-go/blob/31ee46a35443ffd431eeac5005c7c2ed59c79376/core/VERSION

rubenhoenle: This should work for core also: 

`grep -E "(^services/[^/]+/VERSION$|^core/VERSION$)"`

It's even more safe than the current version (on main branch) because it checks for the start of lines also (see the last example below, this would match when using the old pattern)

(+) core/VERSION
(-) core/abc/VERSION
(+) services/cdn/VERSION
(-) services/cdn/sub/VERSION
(-) services/cdn/.openapi-generator/VERSION
(+) services/dns/VERSION
(-) trololol/services/dns/VERSION

marceljk: updated

rubenhoenle (approved):

PR #3548, opened 2025-09-09
fix: add function for proper bytearray to base64 encoding

Benjosh95:
## Description

Jira-ticket: https://jira.schwarz/browse/STACKITSDK-139

## Checklist

- [ ] Issue was linked above
- [ ] Changelogs


core/utils/utils_test.go
```diff
@@ -37,3 +38,122 @@ func TestContainsInt(t *testing.T) {
...
+
+// Test struct for YAML conversion testing
+type TestServer struct {
+	Name     string  `json:"name"`
+	UserData *[]byte `json:"userData,omitempty"`
+	Data     []byte  `json:"data,omitempty"`
+}
+
+func TestConvertByteArraysToBase64(t *testing.T) {
+	// Test with nil input
```
rubenhoenle: Could we have this as a table test maybe? Or is there any reason not to use a table test here?

core/utils/utils.go
```diff
@@ -13,3 +18,50 @@ func Contains[T comparable](slice []T, element T) bool {
...
+	}
+
+	// Convert byte arrays to base64 strings
+	convertByteArraysToBase64Recursive(result)
+
+	return result, nil
+}
+
+// convertByteArraysToBase64Recursive recursively converts []byte values to base64 strings
+func convertByteArraysToBase64Recursive(data interface{}) {
```
rubenhoenle, line 48: could we also have a unit test for this?

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
 		t.Fatalf("Should not be contained")
 	}
 }
+
+// Test struct for YAML conversion testing
+type TestServer struct {
```
rubenhoenle, line 43:
```suggestion
type testServer struct {
```

doesn't need to be exposed I guess

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+		{
+			name:        "nil input",
+			input:       nil,
+			expectError: false,
+		},
+		{
+			name: "normal case with byte arrays",
+			input: TestServer{
+				Name:     "test-server",
+				UserData: func() *[]byte { b := []byte("hello world"); return &b }(),
```
rubenhoenle, line 65:
```suggestion
				UserData: Ptr([]byte("hello world"))
```

https://github.com/stackitcloud/stackit-sdk-go/blob/0b4638e3eb776a9cf5faf422c80c8a0c4c95e7fd/core/utils/utils.go#L3-L6

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+			}
+		}
+	case []interface{}:
+		for _, val := range v {
+			if str := findBase64String(val); str != "" {
+				return str
+			}
+		}
+	case string:
+		if v != "" && v != "hello" && v != "test" && v != "nested" {
```
rubenhoenle, line 198: hardcoded values? :thinking:

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+func TestConvertByteArraysToBase64Recursive(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    interface{}
+		expected string // check if []byte was converted to base64 string
+	}{
+		{"nil", nil, ""},
+		{"map with []byte", map[string]interface{}{"data": []byte("hello")}, "aGVsbG8="},
+		{"slice with []byte", []interface{}{[]byte("test")}, "dGVzdA=="},
+		{"nested map", map[string]interface{}{"level": map[string]interface{}{"data": []byte("nested")}}, "bmVzdGVk"},
```
rubenhoenle, line 162: Please format the input for the table test properly.

```golang
{
  name: "nested map", 
  input: map[string]interface{}{
    "level": map[string]interface{}{
      "data": []byte("nested")
     },
  }, 
  expected: "bmVzdGVk"
},
```

Maybe it's just me, but I'm having a really hard time to get the goal of this PR in it's current state.

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+	}
+}
+
+func TestConvertByteArraysToBase64Recursive(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    interface{}
+		expected string // check if []byte was converted to base64 string
+	}{
+		{"nil", nil, ""},
```
rubenhoenle, line 159: `"nil"`? Accident or intended?

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+		{"map with []byte", map[string]interface{}{"data": []byte("hello")}, "aGVsbG8="},
+		{"slice with []byte", []interface{}{[]byte("test")}, "dGVzdA=="},
+		{"nested map", map[string]interface{}{"level": map[string]interface{}{"data": []byte("nested")}}, "bmVzdGVk"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			convertByteArraysToBase64Recursive(tt.input)
+
+			if tt.expected == "" {
```
rubenhoenle, line 169: why is this skip of the further checks needed?

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+	UserData *[]byte `json:"userData,omitempty"`
+	Data     []byte  `json:"data,omitempty"`
+}
+
+func TestConvertByteArraysToBase64(t *testing.T) {
+	tests := []struct {
+		name           string
+		input          interface{}
+		expectedFields map[string]interface{}
+		expectError    bool
```
rubenhoenle, line 54: You have no test case in your table test where `expectError` is set to true. Then you don't have to have it here.

I would say either add some test cases which are expected to exit with an error or get rid of this input.

core/utils/utils_test.go
rubenhoenle: ...or just move the struct definition into the `TestConvertByteArraysToBase64` func. You're only using it there if I didn't mess something up here. See the example below for reference.

https://github.com/stackitcloud/stackit-cli/blob/a5438f4cac3a794cb95d04891a83252aa9ae1f1e/internal/pkg/utils/strings_test.go#L10-L13

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+				} else if actualValue != expectedValue {
+					t.Fatalf("Expected field %s to be %v, got %v", fieldName, expectedValue, actualValue)
+				}
+			}
+
+			// Check that no unexpected fields exist (except for omitted empty byte arrays)
+			for fieldName, actualValue := range result {
+				if _, expected := tt.expectedFields[fieldName]; !expected {
+					// Allow data field to exist even if not in expectedFields (for empty byte array case)
+					if fieldName == "data" && tt.name == "empty byte array case" {
```
rubenhoenle, line 143: Seems weird to me checking for test case names here. Any reason?

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+			},
+			expectError: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result, err := ConvertByteArraysToBase64(tt.input)
+
+			if tt.expectError {
```
rubenhoenle, line 106:
```suggestion
			if (err != nil) != tt.wantErr {
```

rubenhoenle: This should handle all cases without needing 3 if conditions :smile:

core/utils/utils_test.go
```diff
@@ -37,3 +38,166 @@ func TestContainsInt(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("Expected no error, got: %v", err)
+			}
+
+			// Special case for nil input
+			if tt.input == nil {
+				if result != nil {
+					t.Fatalf("Expected nil result for nil input, got: %v", result)
+				}
+				return
```
rubenhoenle, line 122: Why stop the test case only because the input is nil?

PR #3699, opened 2025-10-29
fix(kms): fix disableKeyVersionWaiter to check for proper state

Benjosh95:
## Description

Jira-Ticket: https://jira.schwarz/browse/STACKITSDK-262

## Checklist

- [x] Issue was linked above
- [x] Changelogs
- [x] VERSION file(s) of the service(s) were adjusted


services/kms/wait/wait.go
```diff
@@ -143,16 +150,36 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
...
 			}
-			return true, nil, err
+			return false, nil, err
+		}
+
+		if response.State != nil {
+			switch *response.State {
+			case kms.VERSIONSTATE_DISABLED:
+				return true, response, nil
+			case kms.VERSIONSTATE_ACTIVE:
```
rubenhoenle:
```suggestion
			case kms.VERSIONSTATE_ACTIVE, VERSIONSTATE_CREATING, VERSIONSTATE_KEY_MATERIAL_UNAVAILABLE:
```

Just combine the cases which have the same effect. Saves you code and is way easier to read

services/kms/wait/wait.go
```diff
@@ -143,16 +150,36 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
...
+				return false, nil, nil
+			case kms.VERSIONSTATE_CREATING:
+				return false, nil, nil
+			case kms.VERSIONSTATE_KEY_MATERIAL_UNAVAILABLE:
+				return false, nil, nil
+			case kms.VERSIONSTATE_DESTROYED:
```
rubenhoenle:
```suggestion
			case kms.VERSIONSTATE_DESTROYED, VERSIONSTATE_KEY_MATERIAL_INVALID:
```

same here

services/kms/wait/wait.go
```diff
@@ -117,6 +118,12 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
 	handler := wait.New(func() (bool, *kms.Version, error) {
 		response, err := client.GetVersionExecute(ctx, projectId, region, keyRingId, keyId, version)
 		if err != nil {
+			var apiErr *oapierror.GenericOpenAPIError
+			if errors.As(err, &apiErr) {
+				if statusCode := apiErr.StatusCode; statusCode == http.StatusNotFound || statusCode == http.StatusGone {
+					return true, nil, fmt.Errorf("enabling failed for key %s version %d: version not found", keyId, version)
```
marceljk: Could be also the case, that the key wasn't found

services/kms/wait/wait.go
```diff
@@ -143,16 +150,36 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
...
 func DisableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, projectId, region, keyRingId, keyId string, version int64) *wait.AsyncActionHandler[kms.Version] {
 	handler := wait.New(func() (bool, *kms.Version, error) {
-		_, err := client.GetVersionExecute(ctx, projectId, region, keyRingId, keyId, version)
+		response, err := client.GetVersionExecute(ctx, projectId, region, keyRingId, keyId, version)
 		if err != nil {
 			var apiErr *oapierror.GenericOpenAPIError
 			if errors.As(err, &apiErr) {
 				if statusCode := apiErr.StatusCode; statusCode == http.StatusNotFound || statusCode == http.StatusGone {
-					return true, nil, nil
+					return true, nil, fmt.Errorf("disabling failed for key %s version %d: version not found", keyId, version)
```
marceljk: same here

services/kms/wait/wait.go
```diff
@@ -143,16 +150,36 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
...
 				}
 			}
-			return true, nil, err
+			return false, nil, err
+		}
+
+		if response.State != nil {
```
marceljk: Nitpick: no blame on you, the same check already exists in this waiter, but here can be a second nil check added for the response. Usually this should never happen, but if so, we can ensure this would break our wait handler.

```go
if response != nil && response.State != nil {
```

When you already work on this waithandler, can you adjust it for the other kms waithandlers as well?

services/kms/wait/wait.go
```diff
@@ -117,16 +118,18 @@ func EnableKeyVersionWaitHandler(ctx context.Context, client ApiKmsClient, proje
...
-		if response.State != nil {
+		if response != nil && response.State != nil {
 			switch *response.State {
-			case kms.VERSIONSTATE_DESTROYED:
-				return true, response, nil
-			case kms.VERSIONSTATE_KEY_MATERIAL_INVALID:
-				return true, response, nil
-			case kms.VERSIONSTATE_DISABLED:
+			case kms.VERSIONSTATE_DESTROYED, kms.VERSIONSTATE_KEY_MATERIAL_INVALID, kms.VERSIONSTATE_DISABLED:
 				return true, response, nil
```
marceljk, line 133: Doesn't really relates to your fix in this PR, but I think this can be quickly fixed as well:
The returns for this states looks a bit wrong to me and was probably set wrong in the past. In my opinion this states should only return an error and not finish the waiter successfully without an error. (Maybe except the disabled state, which the api may could return, if the version was before disabled)

I think the best approach would be here, to only return successfully (without an error) within the waiter, in the specific wanted state and in the default case, an error should be returned.  So that it has a similar approach like you implemented for the DisabledKeyVersionWaitHandler.

PR #3706, opened 2025-11-03
feat: Remove refresh_token grant type from automated refresh

JorTurFer:
SA key flow is designed for machine 2 machine communication, so using `refresh_token` doesn't make sense from operational pov as the sessions doesn't need to be extended (like it happens for users that you want to silent refresh).

Using `refresh_token` grant type it's not needed because for a workload, using the jwt-bearer grant type or refresh token grant type is exactly the same, but SA's `refresh_token` grant type isn't OAuth compliant as it doesn't require `client_id` (which is mandatory). 

I've tested that everything works running the `backgroundrefresh` example 
image

## Checklist

- [x] Changelogs
- [x] VERSION file(s) of the service(s) were adjusted

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -12,7 +12,6 @@
 
 ## v0.18.0
 - **New:** Added duration utils
-- **Chore:** Use `jwt-bearer` grant to get a fresh token instead of `refresh_token`
 
 ## v0.17.3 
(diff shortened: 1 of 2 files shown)


JorTurFer:
Not stale

JorTurFer:
Handled by https://github.com/stackitcloud/stackit-sdk-go/pull/4074

PR #6051, opened 2026-03-30
feat: add missing htt StatusCode and make it settable

mhenselin:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above
- [ x ] **No generated code was adjusted manually** (check [comments in file header](https://github.com/stackitcloud/stackit-sdk-go/blob/783b7fa78e41d072a3ff08e246a13f1bef5aa764/services/dns/api_default.go#L9))
- [ ] Changelogs
    - [ x ] Changelog(s) of the service(s) were adjusted (see e.g. [here](https://github.com/stackitcloud/stackit-sdk-go/blob/f1e375a38064f798821d22a951bc74ca8a9c8845/services/dns/CHANGELOG.md))
- [ x ] VERSION file(s) of the service(s) were adjusted


core/wait/wait.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/stackitcloud/stackit-sdk-go/core/utils"
 )
 
-var RetryHttpErrorStatusCodes = []int{http.StatusBadGateway, http.StatusGatewayTimeout}
+var RetryHttpErrorStatusCodes = []int{http.StatusBadGateway, http.StatusGatewayTimeout, http.StatusServiceUnavailable}
```
cgoetz-inovex, line 13: I've always assumed this is public to allow setting it. With this change I'd be in favour of deprecating this and making it private after the deprecation period.

core/wait/wait.go
```diff
@@ -122,7 +122,7 @@ func (h *AsyncActionHandler[T]) handleError(retryTempErrorCounter int, err error
 		return retryTempErrorCounter, fmt.Errorf("found non-GenericOpenApiError: %w", err)
 	}
 	// Some APIs may return temporary errors and the request should be retried
-	if !utils.Contains(RetryHttpErrorStatusCodes, oapiErr.StatusCode) {
+	if !utils.Contains(h.retryHttpErrorStatusCodes, oapiErr.StatusCode) {
```
cgoetz-inovex, line 133:
```suggestion
	if !slices.Contains(h.retryHttpErrorStatusCodes, oapiErr.StatusCode) {
```
I'd prefer the builtin func instead of the utils one, if we are already touching this line

core/wait/wait_test.go
```diff
@@ -160,7 +160,41 @@ func TestSetTempErrRetryLimit(t *testing.T) {
 			got := New(checkFn)
 			got.SetTempErrRetryLimit(tt.tempErrRetryLimit)
 
-			diff := cmp.Diff(got, want, cmpOpts...)
+			diff := cmp.Diff(want, got, cmpOpts...)
```
cgoetz-inovex, line 164: nice, thx!

core/wait/wait.go
cgoetz-inovex: On the other hand this allows to express: "whatever the default is + sth. else" like:

```
myRetries := append(RetryHttpErrorStatusCodes, http.StatusRequestTimeout)
```

cgoetz-inovex (dismissed):

core/wait/wait.go
```diff
@@ -122,7 +130,7 @@ func (h *AsyncActionHandler[T]) handleError(retryTempErrorCounter int, err error
...
```
cgoetz-inovex, line 133: We'd also need a fallback here when `h.retryHttpErrrorStatusCodes == nil` for people that did not use `New` but a `AsyncActionHandler{}` literal to create an instance.

PR #6789, opened 2026-04-27
Generator: Update SDK /services/telemetrylink

stackit-pipeline:
Generated from GitHub run [25000604884](https://github.com/stackitcloud/stackit-sdk-generator/actions/runs/25000604884)
spec comparison:


services/telemetrylink/v1betaapi/api_default.go
```diff
@@ -0,0 +1,2068 @@
...
+		PatchFolderTelemetryLink Method for PatchFolderTelemetryLink
+
+		Patches the given Telemetry Link within the folder.
+
+		@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+		@param folderId The STACKIT portal folder UUID the resource is located in.
+		@param regionId The STACKIT region name the resource is located in.
+		@return ApiPatchFolderTelemetryLinkRequest
+	*/
+	PatchFolderTelemetryLink(ctx context.Context, folderId string, regionId string) ApiPatchFolderTelemetryLinkRequest
```
Fyusel, line 129: The patch methods should be named "PartialUpdate..."

services/telemetrylink/v1betaapi/api_default.go
```diff
@@ -0,0 +1,2068 @@
...
+		PutFolderTelemetryLink Method for PutFolderTelemetryLink
+
+		Creates or updates the given Telemetry Link within the folder.
+
+		@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+		@param folderId The STACKIT portal folder UUID the resource is located in.
+		@param regionId The STACKIT region name the resource is located in.
+		@return ApiPutFolderTelemetryLinkRequest
+	*/
+	PutFolderTelemetryLink(ctx context.Context, folderId string, regionId string) ApiPutFolderTelemetryLinkRequest
```
Fyusel, line 177: Those "Put" methods should be named "Update"
https://developers.stackit.schwarz/domains/customer-touchpoints/developer-tools/adr/uniformity-layer/#naming-guidelines

Fyusel (changes requested):

services/telemetrylink/v1betaapi/api_default.go
marceljk: In this specific case, it will be "CreateOrUpdate" because based on the description, the Put endpoint is also used for the Createion

services/telemetrylink/v1betaapi/api_default.go
```diff
@@ -0,0 +1,2068 @@
...
+		@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+		@param folderId The STACKIT portal folder UUID the resource is located in.
+		@param regionId The STACKIT region name the resource is located in.
+		@return ApiGetFolderTelemetryLinkRequest
+	*/
+	GetFolderTelemetryLink(ctx context.Context, folderId string, regionId string) ApiGetFolderTelemetryLinkRequest
+
+	// GetFolderTelemetryLinkExecute executes the request
+	//  @return TelemetryLinkResponse
+	GetFolderTelemetryLinkExecute(r ApiGetFolderTelemetryLinkRequest) (*TelemetryLinkResponse, error)
```
marceljk, line 85: At the moment all GET, PUT and PATCH endpoints share the same response schema/struct `TelemetryLinkResponse`. So the response should be everywhere the same. Is this correct?

If you want to change in the future the response of only the `[Action]FolderTelemetryLink` endpoints, this will be a breaking change. Because then you would need to replace the schema, what results here to a new go struct and customers have to make changes.

So I would suggest to have separate response schemas to be in the future more flexible.

Fyusel:
fixed in #7123

PR #7237, opened 2026-05-12
refac(loadbalancer): use new WaiterHelper for waiters

cgoetz-inovex:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above
- [ ] Changelogs


services/loadbalancer/v2api/wait/wait.go
```diff
@@ -23,58 +22,41 @@ const (
...
-	})
+	waitConfig := wait.WaiterHelper[struct{}, string]{
+		FetchInstance: func() (*struct{}, error) {
+			_, err := a.GetLoadBalancer(ctx, projectId, region, instanceId).Execute()
+			return &struct{}{}, err
+		},
+		GetState: func(_ *struct{}) (string, error) {
+			return "", nil
+		},
+	}
```
Fyusel, line 59: What about `DeleteHttpErrorStatusCodes: []int{http.StatusNotFound},`

marceljk, line 59: Error states are missing and in `FetchInstance` the actual instance should be returned.
See example in contribution guide:
https://github.com/stackitcloud/stackit-sdk-go/blob/a1831605701e7a08cbb59bb5e1f42e9f73641645/CONTRIBUTING.md?plain=1#L139-L157

cgoetz-inovex: WaiterHelper sets a default:
https://github.com/stackitcloud/stackit-sdk-go/blob/a1831605701e7a08cbb59bb5e1f42e9f73641645/core/wait/waiterhelper.go#L30

Default includes some more statuses, thought these were ok

cgoetz-inovex: Wouldn't we need to change the return type of `DeleteLoadBalancerWaitHandler` from `*wait.AsyncActionHandler[struct{}]` to `*wait.AsyncActionHandler[loadbalancer.LoadBalancer]` to implement this? Do we accept breaking changes when doing refactorings?

marceljk: Yes this will change the type, but I would say in this case the breaking change is fine. I also did it in the `DeleteProjectWaitHandler` https://github.com/stackitcloud/stackit-sdk-go/blob/5e5af7f2ef88d336dc81d42b445537f56222a2ba/services/resourcemanager/v0api/wait/wait.go#L34

Most users probably ignored the struct{} response because there you get no helpful information, so this breaking change will not affect them. In my opinion, it was not correctly done in the past, and we shouldn't continue with this.
But please mention this breaking change in the changelog

Fyusel: Ah, missed that. I had a look at other waiter implementations and saw that this was mentioned there :see_no_evil:

marceljk (approved):

PR #7362, opened 2026-05-20
feat(iaas): refactor wait handler to use helper struct

marceljk:
## Description

relates to STACKITSDK-374

## Checklist

- [ ] Issue was linked above
- [ ] Changelogs


services/iaas/v2api/wait/wait.go
```diff
@@ -578,6 +466,7 @@ func ProjectRequestWaitHandler(ctx context.Context, a iaas.DefaultAPI, projectId
 }
 
 // AddVolumeToServerWaitHandler will wait for a volume to be attached to a server
+// Deprecated: AddVolumeToServerWaitHandler is deprecated and will be removed after November 2026. Please use instead ProjectRequestWaitHandler
```
marceljk, line 469: I deprecated `AddVolumeToServerWaitHandler` and `RemoveVolumeFromServerWaitHandler` because the implementation is a bit hacky. Instead I suggest to use `ProjectRequestWaitHandler`, where you pass the request ID of the Attach or Detach request and it waits until the operation is done.

services/iaas/v2api/wait/wait.go
```diff
@@ -177,154 +143,116 @@ func ReadyForNetworkAreaDeletionWaitHandler(ctx context.Context, a iaas.DefaultA
...
-		}
-		if server.Id == nil || server.Status == nil {
-			return false, server, fmt.Errorf("create failed for server with id %s, the response is not valid: the id or the status are missing", serverId)
-		}
-		if *server.Id == serverId && *server.Status == ServerActiveStatus {
-			return true, server, nil
-		}
-		if *server.Id == serverId && *server.Status == ErrorStatus {
-			if server.ErrorMessage != nil {
-				return true, server, fmt.Errorf("create failed for server with id %s: %s", serverId, *server.ErrorMessage)
```
cgoetz-inovex, line 322: I've checked some usages of this waiter.
CLI:
```
					_, err = wait.CreateServerWaitHandler(ctx, apiClient, model.ProjectId, model.Region, serverId).WaitWithContext(ctx)
					return err
```
TF:
```
	_, err = wait.CreateServerWaitHandler(ctx, r.client.DefaultAPI, projectId, region, serverId).WaitWithContext(ctx)
	if err != nil {
		core.LogAndAddError(ctx, &resp.Diagnostics, "Error creating server", fmt.Sprintf("server creation waiting: %v", err))
		return
	}
```
In both cases we don't use the returned instance. So currently the information from `server.ErrorMessage` would be lost.

Maybe we could add sth like this to `WaiterHelper`:
```
type WaiterHelper[T any, S comparable] struct {
  // OnErrorState is an optional func to extract a more specific error from an Instance in an  ErrorState
  OnErrorState func(T) error
}

// in func (w *WaiterHelper[T, S]) Wait() AsyncActionCheck[T]
// ...
		// 2. Check if the operation failed
		if slices.Contains(w.ErrorState, state) {
             if w.OnErrorState != nil && instance != nil {
               if err := w.OnErrorState(instance); err != nil {
                  return true, instance, fmt.Errorf("waiting failed. state is %v, err: %w", state, err)
               }
             }
			return true, instance, fmt.Errorf("waiting failed. state is %v", state)
		}
// ...
```
Then we could use this here to report `server.ErrorMessage`

cgoetz-inovex (approved):

services/iaas/v2api/wait/wait.go
marceljk: Yes, I think this is a nice extension. Because usually when a function returns an error, developers assume that the returned response is not set at all and don't read from it. I would suggest to do it in a follow up

marceljk: Created STACKITTPR-668 for that

PR #10169, opened 2026-08-13
Generator: Update SDK /services/kms

stackit-pipeline:
Generated from GitHub run [31696610205](https://github.com/stackitcloud/stackit-api-specifications/actions/runs/31696610205)
spec comparison: https://github.com/stackitcloud/stackit-api-specifications/compare/e9fe931e91e0b8f067d5c12518398a52062ca794...c9c11ba1b198dc952e8e192d1df98d65d85ba977


services/kms/v1api/handwritten.go
```diff
@@ -0,0 +1,31 @@
...
+
+This API provides endpoints for managing keys and key rings.
+
+API version: 1.0.0
+*/
+
+// This is a handwritten file and is not generated.
+
+// Deprecated: Protection is deprecated and will be removed after February 2027. Use instead `string`.
+type Protection = string
```
marceljk, line 14: set Protection to an alias for string. This has the benefit, that calls like kms.Protection("mystring") still work with the updated SDK and the type matches, to configure payloads

marceljk: This is also intentionally only for the new multi API SDK because the old layer is deprecated

SerseusWasTaken (dismissed):

SerseusWasTaken (approved):

PR #10811, opened 2026-08-28
feat(paginate): add paginate package for AIP compliant pagination

cgoetz-inovex:
## Description

relates to #1234

## Checklist

- [ ] Issue was linked above
- [ ] Changelogs


experimental/paginate/paginate.go
```diff
@@ -0,0 +1,189 @@
...
+	Execute() (Resp, error)
+}
+
+type Option func(*options) error
+
+type options struct {
+	pageSize int32
+	limit    *int
+	maxPages *int
+}
```
SerseusWasTaken: A thought: maybe lets call this `config` instead of `options`. 
For example:  the function `applyOptions` below does not really apply the given options to anything, more than anything it collects the given options into the `var cfg options` (which is even called `cfg` here!). If we call the struct `config` in the first place, the mentioned function could be renamed to `func buildConfig(opts []Option) (config, error)` which i think is a more fitting name.

experimental/paginate/paginate.go
```diff
@@ -0,0 +1,189 @@
...
+	for i, opt := range opts {
+		if opt == nil {
+			return options{}, fmt.Errorf("option %d is nil", i+1)
+		}
+		if err := opt(&cfg); err != nil {
+			return options{}, err
+		}
+	}
+	return cfg, nil
+}
```
SerseusWasTaken: Code smell/nitpick: currently the order of which the options are applied matters if the options effect the same field of `options`. I think maybe we should mention that in a comment.

examples/paginate/go.mod
```diff
@@ -0,0 +1,22 @@
+module paginate
```
SerseusWasTaken:
```suggestion
module github.com/stackitcloud/stackit-sdk-go/examples/paginate
```

experimental/paginate/paginate.go
```diff
@@ -0,0 +1,189 @@
...
+	"math"
+)
+
+// Response defines the methods needed for pagination. The result of Request.Execute() must implement this interface.
+type Response[Item any] interface {
+	GetItems() []Item
+	GetNextPageToken() string
+}
+
+// Request defines the methods needed for pagination on te request side.
```
SerseusWasTaken:
```suggestion
// Request defines the methods needed for pagination on the request side.
```

experimental/paginate/paginate.go
```diff
@@ -0,0 +1,189 @@
...
+	return func(o *options) error {
+		if maxPages < 0 {
+			return fmt.Errorf("maximum pages must not be negative: %d", maxPages)
+		}
+		o.maxPages = &maxPages
+		return nil
+	}
+}
+
+// Items returns a lazy iterator over the items of pageable ist operation.
```
SerseusWasTaken:
```suggestion
// Items returns a lazy iterator over the items of pageable list operation.
```

examples/paginate/paginate.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+const region = "eu01"
+
+func main() {
+	ctx := context.Background()
+	projectId := ""
+	if projectId == "" {
+		projectId = os.Getenv("PROJECT_ID")
+	}
```
SerseusWasTaken:
```suggestion
	projectId = os.Getenv("PROJECT_ID")
```

experimental/paginate/paginate.go
```diff
@@ -0,0 +1,189 @@
...
+)
+
+// Response defines the methods needed for pagination. The result of Request.Execute() must implement this interface.
+type Response[Item any] interface {
+	GetItems() []Item
+	GetNextPageToken() string
+}
+
+// Request defines the methods needed for pagination on te request side.
+type Request[Self, Resp any] interface {
```
SerseusWasTaken, line 18: What do you think of changing the signature to `Request[Self, Item any, Resp Response[Item]]`? This way the relationship between `Request` and `Response` is more explicit, and it does not effect any code using the pagination util (examples are unaffected). We could also drop the note in `Response`'s comment: `The result of Request.Execute() must implement this interface.`

SerseusWasTaken (commented):
A few thoughts and typo fixes, but i like it overall. 

One other thing: maybe we should also add a brief README for this package

cgoetz-inovex:
> A few thoughts and typo fixes, but i like it overall.
> 
> One other thing: maybe we should also add a brief README for this package

Not sure what to add to the readme that isn't already in doc comments and examples

SerseusWasTaken (approved):

PR #10842, opened 2026-08-31
feat(UFW): Onboard UFW 

stackit-pipeline:
relates to STACKITSDK-432

Generated from GitHub run [33373337119](https://github.com/stackitcloud/stackit-api-specifications/actions/runs/33373337119)
spec comparison:


services/ufw/v1api/api_default.go
```diff
@@ -0,0 +1,1761 @@
...
+
+		@param ctx context.Context - for authentication, logging, cancellation, deadlines, tracing, etc. Passed from http.Request or context.Background().
+		@param organizationId The STACKIT portal organization UUID
+		@return ApiListOrganizationContainersRequest
+	*/
+	ListOrganizationContainers(ctx context.Context, organizationId string) ApiListOrganizationContainersRequest
+
+	// ListOrganizationContainersExecute executes the request
+	//  @return ListContainersResponse
+	ListOrganizationContainersExecute(r ApiListOrganizationContainersRequest) (*ListContainersResponse, error)
```
marceljk, line 104: ListFolderContainers and ListOrganizationContainers return both the same response struct. Not sure if this is intended. If one of both needs to be adjusted in the future, they need to do a breaking change. I would suggest to use separate schemas for both, except they are 100% sure they will remain always the same

alextiugan15: Thanks for pointing this up. 

Yeah, sharing schemas can definitely be risky if the endpoints end up needing different fields later.For this specific case, I went with a shared ListContainersResponse because a container is represented exactly the same way whether it's under a folder or an organization. Since both endpoints are just listing children, I'm really confident the response structures will stay perfectly aligned.  

Keeping them combined keeps the spec clean and avoids duplicating things prematurely. If we ever do hit an edge case down the road where they need to diverge, we can always split them into separate schemas then. Let me know if you still feel strongly about separating them right now, happy to chat more about it in that case.

marceljk: I totally agree with you that summarizing it to one schema makes things easier. But if we do hit the edge case, and you split the schemas, it results in a breaking change within the SDK for all our customers, which we always try to avoid. If you are confident that there will be no edge case in the future or if the split would happen then in a new API version, I'm also fine with leaving it now like it is.

alextiugan15: I’m very confident that the core container structure won’t diverge between folders and organizations, which is why I'd still lean toward keeping them shared for now. 

However, I completely agree with your condition, if we ever do hit an edge case where they need different fields, we will handle it by bumping the API version rather than breaking the current v1 SDK.  Thanks for talking this through with me. I'll leave it as is with that plan in mind.

GokceGK (approved):

Fyusel (approved):

Issue #11084, opened 2026-09-04
core/wait: waiter returns (nil, nil) on a retryable 502/504 and never retries


devpie:
## Description

A wait handler built with `wait.WaiterHelper` reports **success with a nil response** when a single poll hits a
retryable HTTP status (502 or 504), and the retry it is supposed to perform never happens.

Two pieces of `core/wait` disagree about what `done` means:

`WaiterHelper.Wait()` returns `true` for `waitFinished` on *any* fetch error, so the error is the only thing that
distinguishes "finished" from "broken"
([waiterhelper.go](https://github.com/stackitcloud/stackit-sdk-go/blob/core/v0.27.0/core/wait/waiterhelper.go)):

```go
instance, err := w.FetchInstance()
if err != nil {
    ...
    return true, nil, err
}
```

`WaitWithContext` then hands that error to `handleError`, which **swallows** a retryable status and returns a nil
error — but the `if done` check that follows does not know the error was swallowed
([wait.go](https://github.com/stackitcloud/stackit-sdk-go/blob/core/v0.27.0/core/wait/wait.go)):

```go
done, res, err := h.checkFn()
if err != nil {
    retryTempErrorCounter, err = h.handleError(retryTempErrorCounter, err)
    if err != nil {
        return res, err
    }
}
if done {
    return res, nil   // res is nil, err was swallowed, no retry
}
```

So the caller gets `(nil, nil)`. Callers reasonably treat a nil error as success and dereference the response, which
panics. `tempErrRetryLimit` is never reached because control never returns to the loop.

## Steps to reproduce

```golang
package main

import (
	"context"
	"fmt"
	"net/http"
	"time"

	"github.com/stackitcloud/stackit-sdk-go/core/oapierror"
	"github.com/stackitcloud/stackit-sdk-go/core/wait"
)

type instance struct{ Name string }

func main() {
	calls := 0

	helper := wait.WaiterHelper[instance, string]{
		// One single 502 from the API, then the resource would be ACTIVE.
		FetchInstance: func() (*instance, error) {
			calls++
			if calls == 1 {
				return nil, &oapierror.GenericOpenAPIError{StatusCode: http.StatusBadGateway}
			}
			return &instance{Name: "my-resource"}, nil
		},
		GetState:    func(i *instance) (string, error) { return "ACTIVE", nil },
		ActiveState: []string{"ACTIVE"},
		ErrorState:  []string{"ERROR"},
	}

[...]

PR #11418, opened 2026-09-16
fix(core): fix waiter bug that caused it to return nil nil

SerseusWasTaken:
## Description
Fixes a bug in waithandlerhelper that caused some wait handlers to return (nil, nil).

It also introduces a way to prevent future wait handlers to return the wrong `waitFinished` state by overriding it to `false` in wait.go. The related regression test proves that the workaround works, while the other `TestWaiterHelper_WaitWithContext` test covers more of the waiter helper handler implementation.

Relates to STACKITTPR-844, https://github.com/stackitcloud/stackit-sdk-go/issues/11084 and should fix downstream issue https://github.com/stackitcloud/terraform-provider-stackit/issues/1743 once used in the provider

## Checklist

- [x] Issue was linked above
- [x] Changelogs


core/wait/wait.go
```diff
@@ -102,6 +102,10 @@ func (h *AsyncActionHandler[T]) WaitWithContext(ctx context.Context) (res *T, er
 			if err != nil {
 				return res, err
 			}
+			// the error was retryable and got swallowed by h.handleError, so done represents a failed
+			// fetch rather than a finished action - poll again instead of returning
+			// otherwise we might return (nil, nil)
+			done = false
```
SerseusWasTaken, line 108: This guard is debatable

core/wait/wait_test.go
```diff
@@ -386,6 +386,46 @@ func TestWaitWithContext(t *testing.T) {
...
+			t.Errorf("expected no error, got %v", err)
+		}
+		if resp == nil || resp.Name != "my-resource" {
+			t.Errorf("expected a resolved response, got %v", resp)
+		}
+		if numberCheckFnCalls != 2 {
+			t.Errorf("expected checkFn to be called twice (initial + retry), got %d calls", numberCheckFnCalls)
+		}
+	})
+}
```
SerseusWasTaken, line 427: This regression test is not needed if we decide to get rid of the guard

core/wait/waiterhelper.go
```diff
@@ -44,7 +44,7 @@ func (w *WaiterHelper[T, S]) Wait() AsyncActionCheck[T] {
 					return true, nil, nil
 				}
 			}
-			return true, nil, err
+			return false, nil, err
 		}
 
 		state, err := w.GetState(instance)
```
marceljk, line 50: I can't mark here the correct line, but I would suggest to change the return in the error handling of GetState as well, to return there also false for waitFinished.
Usually it wouldn't return an openapi error, so this change has no impact, but maybe there will be some edge cases in the future which will be affected by this

core/wait/wait.go
marceljk: For me it feels a bit wrong to change a return value of a function within the error handling. I think the better approach would be, to change the `if done` condition to an `else if done` condition.
With this approach we don't need to change a return value of the function and would only take care of the return value, if there is no error.

core/wait/waiterhelper_test.go
```diff
@@ -161,3 +164,166 @@ func TestWaiterHelper_Wait(t *testing.T) {
...
+		fetchResponses             []fetchResponse
+		activeStates               []string
+		errorStates                []string
+		deleteHttpErrorStatusCodes []int
+		wantCalls                  int
+		wantErr                    bool
+		wantResponse               *MockResource
+	}{
+		{
+			name: "Success - Retryable 502 Gateway Error followed by Active State",
```
marceljk, line 186: For completeness we could also add a test to check the 504 error

core/wait/wait.go
SerseusWasTaken: Unless i'm missing something, i don't think we can do that. `waiterhelper.go` intentionally returns true and an error:
```go
// 2. Check if the operation failed
if slices.Contains(w.ErrorState, state) {
  return true, instance, fmt.Errorf("waiting failed. state is %v", state)
}

```
this signals that wait is done and there was an error state. I think this is a valid, as waiting can be done AND there can be an error state.
The guard fixes the edge case because a few lines above `h.handleError` swallows the retryable error

core/wait/waiterhelper.go
SerseusWasTaken: Good point, will change

core/wait/wait.go
marceljk: I think the behavior will remain the same with the `else if done`.


> Unless i'm missing something, i don't think we can do that. `waiterhelper.go` intentionally returns true and an error:
> 
> ```go
> // 2. Check if the operation failed
> if slices.Contains(w.ErrorState, state) {
>   return true, instance, fmt.Errorf("waiting failed. state is %v", state)
> }
> ```

When the mentioned error will be returned, the WaitWithContext function would end anyways with an error here, independent if waitFinished is true or false. The `handleError()` function will return an error because it can't parse the error to a `GenericOpenAPIError`.

https://github.com/stackitcloud/stackit-sdk-go/blob/6030ec192deab9e3e8ac4eab0dace31b081f3625/core/wait/wait.go#L99-L104

SerseusWasTaken: I think you're right. My regression test still passes, so i think this is safe. I will push the change.

marceljk (approved):

